PR #22, opened 2022-12-13
MF-21 - Read all users with no limit

zzokki81:
Read all users with no limit. Use -1 query param.


users/mocks/users.go
```diff
@@ -111,11 +111,18 @@ func (urm *userRepositoryMock) RetrieveAll(ctx context.Context, offset, limit ui
 		return up, nil
 	}
 
-	for _, u := range sortUsers(mockUsers) {
-		if i >= offset && i < (limit+offset) {
+	sortedUsers := sortUsers(mockUsers)
+	for _, u := range sortedUsers {
+		switch limit {
```
manuio: You can probably simplify it with an if case

manuio (changes requested):

manuio (approved):
LGTM

PR #35, opened 2023-02-13
MF-9 Fix delete endpoint. Check owner for delete things.

zzokki81:
Fix delete endpoint. Check owner for delete things.


manuio (approved):
LGTM

PR #47, opened 2023-02-23
MF-40 - Add check for invalid id removal.

### LINKED ISSUE #40: Fix things and channels removal for invalid IDs
When trying to delete a thing or a channel there is no error returned if the ID isn't found. Fix it and return 404 NotFound.

zzokki81:
Add check for invalid thing and channel id removal. Fixed tests. Fixed mocks.
Resolves #40


manuio (approved):
LGTM

PR #63, opened 2023-03-01
NOISSUE - Fix RetrieveByOwner where clause for groups and orgs

zzokki81:
Fix `RetrieveByOwner` where clause for `groups` and `orgs`


manuio (approved):
LGTM

PR #101, opened 2023-03-29
MF-93 - Implement Access Control to View Things

### LINKED ISSUE #93: Implement Access Control to View Things
ViewThing should be accessible by owner or members of groups where the Thing is assigned

zzokki81:
Implement check is user  group owner or member for `ViewThing` 

Resolves #93


things/service.go
```diff
@@ -252,6 +252,20 @@ func (ts *thingsService) ViewThing(ctx context.Context, token, id string) (Thing
...
+
+	mpg, err := ts.groups.RetrieveMemberships(ctx, res.GetId(), PageMetadata{})
+	if err != nil {
+		return Thing{}, err
+	}
+
+	if len(opg.Groups) == 0 && len(mpg.Groups) == 0 {
+		return Thing{}, errors.ErrNotFound
+	}
+
```
manuio: What we want is to check if user is member of the group where the thing is assigned. Here you are checking owned groups and membershops with user ID. This is not good.

manuio (changes requested):

manuio (approved):
LGTM

PR #106, opened 2023-03-31
NOISSUE - Fix validate list subscriptions request.

zzokki81:
Fixed token missing error when only key is provided. Validation now checking  both,  token and key.


manuio (approved):
LGTM

PR #125, opened 2023-04-26
MF-124 - Check user permissions in Org according to their roles

zzokki81:
Added methods in Org service that will check permissions against the user role.

### DIFF
--- a/auth/service.go
+++ b/auth/service.go
@@ -270,7 +270,7 @@ func (svc service) RemoveOrg(ctx context.Context, token, id string) error {
 		return err
 	}
 
-	if err := svc.canAccessOrg(ctx, id, user.ID); err != nil {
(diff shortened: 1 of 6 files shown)


auth/service.go
```diff
@@ -308,13 +308,13 @@ func (svc service) ViewOrg(ctx context.Context, token, id string) (Org, error) {
 		return Org{}, err
 	}
 
-	org, err := svc.orgs.RetrieveByID(ctx, id)
-	if err != nil {
+	if err := svc.canView(ctx, id, user.ID); err != nil {
```
manuio: Rename it to `canAccessOrg`

manuio (changes requested):

manuio (approved):
LGTM

PR #132, opened 2023-05-04
MF-131 - Fix CanAccessGroup org method

### LINKED ISSUE #131: Improve  CanAccessGroup org method
`CanAccessGroup`  is used to retrieve `orgs` where group is assigned. Improve this method and make it more readable.

zzokki81:
Fixed `CanAccessGroup` to check if there is `orgs` where group is assigned.

Resolves #131

### DIFF
--- a/auth/service.go
+++ b/auth/service.go
@@ -604,8 +604,7 @@ func (svc service) CanAccessGroup(ctx context.Context, token, groupID string) er
 	}
 
 	for _, org := range op.Orgs {
-		err := svc.canEditGroups(ctx, org.ID, user.ID)
-		if err == nil {
+		if err := svc.canEditGroups(ctx, org.ID, user.ID); err == nil {
 			return nil
 		}
 	}
--- a/auth/service.go
+++ b/auth/service.go
@@ -603,13 +603,17 @@ func (svc service) CanAccessGroup(ctx context.Context, token, groupID string) er
 		return err
 	}
 
+	if len(op.Orgs) == 0 {
+		return errors.ErrAuthorization
+	}
+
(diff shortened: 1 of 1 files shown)


auth/service.go
```diff
@@ -603,13 +603,17 @@ func (svc service) CanAccessGroup(ctx context.Context, token, groupID string) er
 		return err
 	}
 
+	if len(op.Orgs) == 0 {
```
manuio: The length check is going to be done in the for loop, if len(op.Orgs) is 0 it will be ignored. Better move the error at the end of the method

auth/service.go
```diff
@@ -603,13 +603,17 @@ func (svc service) CanAccessGroup(ctx context.Context, token, groupID string) er
...
+		return errors.ErrAuthorization
+	}
+
 	for _, org := range op.Orgs {
 		if err := svc.canEditGroups(ctx, org.ID, user.ID); err != nil {
```
manuio: `err == nil` The point is to know if it's ok for 1 org

manuio (changes requested):

auth/service.go
```diff
@@ -604,8 +604,9 @@ func (svc service) CanAccessGroup(ctx context.Context, token, groupID string) er
 	}
 
 	for _, org := range op.Orgs {
-		if err := svc.canEditGroups(ctx, org.ID, user.ID); err != nil {
-			return err
+		err := svc.canEditGroups(ctx, org.ID, user.ID)
```
manuio: inline

manuio (changes requested):

manuio (approved):
LGTM

PR #135, opened 2023-05-05
NOISSUE - Remove unnecessary access org check

zzokki81:
Removed unnecessary check if the member can access the organization.


manuio (approved):
LGTM

PR #151, opened 2023-05-15
MF-150 - Proper delete influxDB bucket

### LINKED ISSUE #150: Proper delete influxDB bucket
In the `reader'  influx tests, the data from the bucket was not cleaned properly before running the tests. When a bucket is deleted, data is still available for a period of time causing tests to sometimes fail.

zzokki81:
Add proper delete `influxDB` bucket in readers test.

Resolves #150


manuio (approved):
LGTM

PR #152, opened 2023-05-16
MF-139 - Fix returned error when adding non existent users

### LINKED ISSUE #139: Fix returned error when adding non existent users
When adding non existent users the error message returned is `{"error":"missing or invalid bearer user token"}`. Fix it to return NotFound error with proper message `User email not found`

zzokki81:
Return  `not found` instead of `internal server error` when adding non existent users

Resolves #139


manuio (approved):
LGTM

PR #170, opened 2023-05-25
MF-169 - Remove all org and group members before delete org.

### LINKED ISSUE #169: Remove all org and group members before delete org
Deleting Org will result  with an error.  Remove all org and group members before delete org

zzokki81:
Removed all  org and group members before delete org.

resolves #169


manuio (approved):
LGTM

PR #186, opened 2023-06-08
MF-185 - Add a check if the user is the owner of the thing in UpdateThing service

### LINKED ISSUE #185: Fix update things rights
Currently any valid token can update the thing. Fix it so that only the owner has update rights
[Link](https://github.com/MainfluxLabs/mainflux/blob/02d469b2bab510c96c939fd0219bba12e5ce81c0/things/service.go#L228)

zzokki81:
Added check is user owner of the thing in `UpdateThing`  service method
Resolves #185

### DIFF
--- a/things/service.go
+++ b/things/service.go
@@ -231,7 +231,9 @@ func (ts *thingsService) UpdateThing(ctx context.Context, token string, thing Th
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


manuio (approved):
LGTM

PR #206, opened 2023-06-14
MF-205 - Add permission to admin to list all groups

### LINKED ISSUE #205: Add admin permission to list groups.
Currently only user can list groups. Add permission to the administrator so he can  list them too.

zzokki81:
Added permission to admin to list all groups.

Resolves #205

### DIFF
--- a/things/postgres/groups.go
+++ b/things/postgres/groups.go
@@ -11,11 +11,10 @@ import (
 	"strings"
 	"time"
 
-	"github.com/jackc/pgerrcode"
-	"github.com/jackc/pgx/v5/pgconn"
-
 	"github.com/MainfluxLabs/mainflux/pkg/errors"
 	"github.com/MainfluxLabs/mainflux/things"
+	"github.com/jackc/pgerrcode"
+	"github.com/jackc/pgx/v5/pgconn"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


things/postgres/groups.go
```diff
@@ -523,11 +463,90 @@ func (gr groupRepository) UnassignMember(ctx context.Context, groupID string, id
...
 
-type dbMember struct {
-	MemberID  string    `db:"member_id"`
-	GroupID   string    `db:"group_id"`
-	CreatedAt time.Time `db:"created_at"`
-	UpdatedAt time.Time `db:"updated_at"`
+func (gr groupRepository) retrieve(ctx context.Context, ownerID string, pm things.PageMetadata) (things.GroupPage, error) {
+	var ownq string
+	if ownerID != "" {
+		ownq = "owner_id = :owner_id"
```
manuio, line 467: Can we use `getOwnerQuery` here?

manuio (changes requested):

things/postgres/groups.go
zzokki81: I don't think so.  In group SQL table we have `owner_id` and in channel SQL tale `owner`

manuio: Right!

manuio (approved):
LGTM

manuio (approved):
LGTM

PR #213, opened 2023-06-22
MF-211 - Add Admin query parameter to list or not as admin

### LINKED ISSUE #211: Add Admin query parameter to list or not as admin
Add a query parameter in Things and Auth service to list Orgs, Groups,  Channels and Things as owner (admin) or admin. When the parameter `admin=true` is present in the query we list all the entities. When the parameter is not present in the query we list the entities like for any other owner in the platform.

zzokki81:
Added query parameter `admin` which will enable to list  orgs, things, channels and groups fro all users.
If the administrator sends a request without `admin` query parameter , the data will be listed as a regular user.

Resolves #211


manuio (approved):
LGTM

PR #218, opened 2023-06-23
MF-217 - Add admin permission to list channel messages of all users

### LINKED ISSUE #217: Add to admin permission to list channel messages
Add admin permission to list channel messages from each user.

zzokki81:
Added admin permission to list channel messages of all users

Resolves #217

### DIFF
--- a/readers/api/transport.go
+++ b/readers/api/transport.go
@@ -40,10 +40,8 @@ const (
 )
 
 var (
-	errThingAccess = errors.New("thing has no permission")
-	errUserAccess  = errors.New("user has no permission")
-	things         mainflux.ThingsServiceClient
-	auth           mainflux.AuthServiceClient
+	things mainflux.ThingsServiceClient
+	auth   mainflux.AuthServiceClient
 )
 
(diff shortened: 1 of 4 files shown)


readers/api/transport.go
```diff
@@ -325,17 +323,14 @@ func authorize(ctx context.Context, token, key, chanID string) (err error) {
...
 		}
+
+		if err := authorizeAdmin(ctx, user.Email); err == nil {
+			return nil
+		}
+
 		if _, err = things.IsChannelOwner(ctx, &mainflux.ChannelOwnerReq{Owner: user.Id, ChanID: chanID}); err != nil {
-			e, ok := status.FromError(err)
-			if ok && e.Code() == codes.PermissionDenied {
-				return errors.Wrap(errUserAccess, err)
```
manuio, line 337: If `errUserAccess` is not used anymore remove the error. Same for `errThingAccess`. Also, check that both error can be distinguished and add the corresponding tests.

manuio (changes requested):

manuio (approved):
LGTM

PR #222, opened 2023-06-27
MF- 221 - Add update admin in restore endpoint

### LINKED ISSUE #221: Add update admin in restore endpoint
After performing a backup of the database, cleaning volumes, and restoring the backup, the `ownerID` for the admin user gets changed. This change in `ownerID` is causing an incorrect view of channels and groups in the admin page.

To address this issue, we need to make the following modifications in the restore endpoint and related service:
Add the admin user as a request parameter in the restore endpoint.
Update the service to use the old admin user ID and email during the restore process.

zzokki81:
Added update admin in restore endpoint.

Resolves #221

### DIFF
--- a/users/api/endpoint.go
+++ b/users/api/endpoint.go
@@ -235,7 +235,7 @@ func backupEndpoint(svc users.Service) endpoint.Endpoint {
 			return nil, err
 		}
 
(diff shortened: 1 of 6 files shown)


manuio (approved):
LGTM

PR #225, opened 2023-06-28
MF-224 - Fix backup response and restore request in things service

### LINKED ISSUE #224: Fix backup response and restore request in things service
Fix `backup` response and `restore` request in things service.

zzokki81:
Fixed backup response and restore request

Resolves #224

### DIFF
--- a/things/api/things/http/responses.go
+++ b/things/api/things/http/responses.go
@@ -60,7 +60,6 @@ func (res shareThingRes) Empty() bool {
 type thingRes struct {
 	ID       string                 `json:"id"`
 	Name     string                 `json:"name,omitempty"`
-	Owner    string                 `json:"-"`
 	Key      string                 `json:"key"`
 	Metadata map[string]interface{} `json:"metadata,omitempty"`
 	created  bool
@@ -127,7 +126,6 @@ func (res thingsPageRes) Empty() bool {
 type channelRes struct {
 	ID       string                 `json:"id"`
 	Name     string                 `json:"name,omitempty"`
-	Owner    string                 `json:"-"`
 	Metadata map[string]interface{} `json:"metadata,omitempty"`
 	created  bool
 }
--- a/things/api/things/http/responses.go
+++ b/things/api/things/http/responses.go
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	"github.com/MainfluxLabs/mainflux"
-	"github.com/MainfluxLabs/mainflux/things"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


things/postgres/channels.go
```diff
@@ -488,7 +488,7 @@ func (cr channelRepository) retrieve(ctx context.Context, owner string, pm thing
 		olq = ""
 	}
 
-	q := fmt.Sprintf(`SELECT id, name, metadata FROM channels %s ORDER BY %s %s %s;`, whereClause, oq, dq, olq)
+	q := fmt.Sprintf(`SELECT id, name, owner, metadata FROM channels %s ORDER BY %s %s %s;`, whereClause, oq, dq, olq)
```
manuio: Wait, we don't have to retrieve owner all the time. For example, for method `RetrieveByOwner` we don't need it.
Revert this and create a way to do it by passing an argument to `retrieve` method. Eventually, why not `SELECT * FROM channels;`?

manuio (changes requested):

things/postgres/things.go
```diff
@@ -400,7 +400,7 @@ func (tr thingRepository) retrieve(ctx context.Context, owner string, pm things.
 		olq = ""
 	}
 
-	q := fmt.Sprintf(`SELECT id, name, key, metadata FROM things %s ORDER BY %s %s %s;`, whereClause, oq, dq, olq)
+	q := fmt.Sprintf(`SELECT id, name, owner, key, metadata FROM things %s ORDER BY %s %s %s;`, whereClause, oq, dq, olq)
```
manuio: Same here

things/api/things/http/responses.go
```diff
@@ -61,6 +60,7 @@ func (res shareThingRes) Empty() bool {
 type thingRes struct {
 	ID       string                 `json:"id"`
 	Name     string                 `json:"name,omitempty"`
+	Owner    string                 `json:"-"`
```
manuio: This should probably be removed

things/api/things/http/responses.go
```diff
@@ -127,6 +127,7 @@ func (res thingsPageRes) Empty() bool {
 type channelRes struct {
 	ID       string                 `json:"id"`
 	Name     string                 `json:"name,omitempty"`
+	Owner    string                 `json:"-"`
```
manuio: same here

manuio (changes requested):

manuio (approved):
LGTM

PR #226, opened 2023-06-28
NOISSUE - Fix update admin id

zzokki81:
Fixed query for update admin ID. Remove unused code.

### DIFF
--- a/users/postgres/users.go
+++ b/users/postgres/users.go
@@ -81,7 +81,7 @@ func (ur userRepository) Update(ctx context.Context, user users.User) error {
 }
 
 func (ur userRepository) UpdateUser(ctx context.Context, user users.User) error {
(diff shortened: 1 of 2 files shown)


users/postgres/users.go
```diff
@@ -81,7 +81,7 @@ func (ur userRepository) Update(ctx context.Context, user users.User) error {
 }
 
 func (ur userRepository) UpdateUser(ctx context.Context, user users.User) error {
-	q := `UPDATE users SET metadata = :metadata WHERE email = :email AND status = 'enabled'`
+	q := `UPDATE users SET id = :id, metadata = :metadata WHERE email = :email AND status = 'enabled'`
```
manuio, line 84: This should be removed after to implement this feature: https://github.com/MainfluxLabs/mainflux/issues/234

PR #229, opened 2023-06-30
MF-223 - Add users http api package

### LINKED ISSUE #223: Create users http api package
Create new package using files from here: https://github.com/MainfluxLabs/mainflux/tree/master/users/api

zzokki81:
Created users http api package.

Resolves #223


manuio (approved):
LGTM

PR #242, opened 2023-07-10
MF-234 - Add users roles

### LINKED ISSUE #234: Cretae Admin roles in auth service
As a first solution we decided to use a harcoded admin email as authorization while we were implementing organizations and policies. This is not good for many reasons and must be replaced. 

Create users_roles table in auth service with `super_admin` and `admin` roles. `users_roles` table will have 2 columns: `role` and `user_id`. For operations that need admin rights we will check during the authorization if the user ID corresponds to roles `root_:admin` or `admin`. The DB layer must allow to create, read, update and delete a role. Delete can only be done if the user is disabled and if user is not `root_admin`.

Finally, we will add the role in the View user response in order to use it from applications to display certain pages. 

When this issue will be fixed remove in UpdateUsers method the possibility to change the user ID added in this PR: https://github.com/MainfluxLabs/mainflux/pull/226/files

zzokki81:
Created users roles for authorization.

Resolves: #234


auth.proto
```diff
@@ -137,3 +138,8 @@ message GroupsReq {
 message GroupsRes {
     repeated Group groups = 1;
 }
+
+message SaveRoleReq {
```
manuio: CreateRoleReq

auth/api/grpc/client.go
```diff
@@ -29,6 +29,7 @@ type grpcClient struct {
 	canAccessGroup endpoint.Endpoint
 	assign         endpoint.Endpoint
 	members        endpoint.Endpoint
+	saveRole       endpoint.Endpoint
```
manuio: createRole

auth/api/grpc/client.go
```diff
@@ -83,6 +84,14 @@ func NewClient(tracer opentracing.Tracer, conn *grpc.ClientConn, timeout time.Du
 			decodeMembersResponse,
 			mainflux.MembersRes{},
 		).Endpoint()),
+		saveRole: kitot.TraceClient(tracer, "save_role")(kitgrpc.NewClient(
```
manuio: createRole and create_role

auth/api/grpc/client.go
```diff
@@ -83,6 +84,14 @@ func NewClient(tracer opentracing.Tracer, conn *grpc.ClientConn, timeout time.Du
...
+			conn,
+			svcName,
+			"SaveRole",
```
manuio: CreateRole

manuio (changes requested):

auth/roles.go
```diff
@@ -0,0 +1,27 @@
+package auth
+
+import "context"
+
+const (
+	// RoleAdmin is the admin role.
+	RoleAdmin = "admin"
+
+	// RoleSuperAdmin is the super admin role.
```
manuio: RoleRootAdmin

manuio: Define it first

manuio (changes requested):

cmd/auth/main.go
```diff
@@ -314,10 +314,13 @@ func newService(db *sqlx.DB, tc mainflux.ThingsServiceClient, uc mainflux.UsersS
...
 	keysRepo := tracing.New(postgres.New(database), tracer)
 
+	roleRepo := postgres.NewRoleRepo(db)
+	roleRepo = tracing.RoleRepositoryMiddleware(tracer, roleRepo)
+
 	idProvider := uuid.New()
 	t := jwt.New(secret)
 
-	svc := auth.New(orgsRepo, tc, uc, keysRepo, idProvider, t, duration, adminEmail)
+	svc := auth.New(orgsRepo, tc, uc, keysRepo, roleRepo, idProvider, t, duration, adminEmail)
```
manuio: Remove the adminEmail

manuio (changes requested):

auth/api/http/orgs/endpoint_test.go
```diff
@@ -1669,6 +1673,7 @@ func TestRestore(t *testing.T) {
 	_, viewerToken, err := svc.Issue(context.Background(), "", auth.Key{Type: auth.LoginKey, IssuedAt: time.Now(), IssuerID: viewerID, Subject: viewerEmail})
 	assert.Nil(t, err, fmt.Sprintf("Issuing login key expected to succeed: %s", err))
 
+
```
manuio: why?

auth/roles.go
```diff
@@ -0,0 +1,27 @@
...
+	RoleRootAdmin = "root_admin"
+)
+
+type RoleService interface {
```
manuio: Role or Roles?

auth/roles.go
```diff
@@ -0,0 +1,27 @@
...
+	// SaveRole saves the user role.
+	SaveRole(ctx context.Context, id, role string) error
+}
+
+type RoleRepository interface {
```
manuio: Same here

manuio (changes requested):

auth/postgres/roles.go
```diff
@@ -0,0 +1,87 @@
...
+
+	if _, err := rr.db.NamedExecContext(ctx, q, dbur); err != nil {
+		return errors.Wrap(errors.ErrRemoveEntity, err)
+	}
+
+	return nil
+}
+
+type dbUserRole struct {
+	ID   string `db:"user_id"`
```
manuio: Field name must follow the db name. Better `UserID`

manuio (changes requested):

auth/api/logging.go
```diff
@@ -327,3 +327,15 @@ func (lm *loggingMiddleware) Restore(ctx context.Context, token string, backup a
 
 	return lm.svc.Restore(ctx, token, backup)
 }
+
+func (lm *loggingMiddleware) AssignRole(ctx context.Context, id, role string) (err error) {
+	defer func(begin time.Time) {
+		message := fmt.Sprintf("Method save_role for id %s and role %s took %s to complete", id, role, time.Since(begin))
```
manuio: `assign_role`

manuio (changes requested):

auth/postgres/roles.go
```diff
@@ -0,0 +1,87 @@
...
+
+	return nil
+}
+
+func (rr rolesRepository) RetrieveRole(ctx context.Context, userID string) (string, error) {
+	q := `SELECT role FROM users_roles WHERE user_id = $1;`
+
+	dbur := dbUserRole{UserID: userID}
+
+	if err := rr.db.QueryRowxContext(ctx, q, userID).StructScan(&dbur); err != nil {
```
manuio: Use `NamedExecContext` instead of `QueryRowxContext`. It's more readable

auth/postgres/roles.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/MainfluxLabs/mainflux/pkg/errors"
+)
+
+var _ auth.RolesRepository = (*rolesRepository)(nil)
+
+type rolesRepository struct {
+	db Database
+}
+
+// NewRolesRepo instantiates a PostgreSQL implementation of roles
```
manuio: This comment can be done in a single line

auth.proto
```diff
@@ -71,7 +72,7 @@ message IssueReq {
 }
 
 message AuthorizeReq {
-    string email = 1;
+    string token = 1;
```
manuio: This is ok but I forgot something, we need more fields :) `object`, `subject` and `action`.   Subjects can be `system`, `orgs`,  `groups`, `things` and `channels`. Object is the ID of the entity were we want to access. Actions are `read`, `write`, `execute` and their combinations. I'll open an issue about but you can already add `object`, `subject` and `action` fields.

auth/service.go
```diff
@@ -168,9 +169,14 @@ func (svc service) Identify(ctx context.Context, token string) (Identity, error)
 	}
 }
 
-func (svc service) Authorize(ctx context.Context, pr AuthzReq) error {
-	if pr.Email != svc.adminEmail {
-		return errors.ErrAuthorization
+func (svc service) Authorize(ctx context.Context, ar AuthzReq) error {
```
manuio, line 159: Here we need a switch case depending on the `Subject`. If the subject is `root` then we check if the token correspond to user `root_admin`. You can ignore any other subject for now. Something like that:
```
user, err := svc.identify(ctx, ar.Token)
if err != nil {
	return err
}

switch ar.Subjsect {
    case rootSubject:
	return svc.rootAction(ctx, user.ID)
    case groupSubject:
       return svc.groupAction(ctx, ar.Object, ar.Action, user.ID)
    default:
       return  errors.New("Unknwon Subject")
}

    return nil
```

auth/service.go
```diff
@@ -168,9 +169,14 @@ func (svc service) Identify(ctx context.Context, token string) (Identity, error)
...
+	user, err := svc.Identify(ctx, ar.Token)
```
manuio: If you cal `svc.Identity` it's going to be logged while `Authorize` it's already creating a log. Create a helper method `identity` and call it from `Identity` and `Authorize` to avoid the double log.

auth/service.go
```diff
@@ -281,7 +287,7 @@ func (svc service) ListOrgs(ctx context.Context, token string, admin bool, pm Pa
 	}
 
 	if admin {
-		if err := svc.Authorize(ctx, AuthzReq{Email: user.Email}); err == nil {
+		if err := svc.isAdmin(ctx, user.ID); err == nil {
```
manuio: Revert to `svc.Authorize(ctx, AuthzReq{Subject: "root", Token: token, Action: ""l})`

things/mocks/auth.go
```diff
@@ -41,11 +41,15 @@ func (svc authServiceMock) Issue(ctx context.Context, in *mainflux.IssueReq, opt
 }
 
 func (svc authServiceMock) Authorize(ctx context.Context, req *mainflux.AuthorizeReq, _ ...grpc.CallOption) (r *empty.Empty, err error) {
-	if req.GetEmail() != "admin@example.com" {
-		return &empty.Empty{}, errors.ErrAuthorization
+	if req.GetToken() == "adminToken" {
```
manuio, line 44: Here you have to copy the same switch case than in the service layer but instead of saving and reading from postgres you do it from `svc.sysRoles`, `svc.groupRoles`, `thingsRoles` etc etc For now you can create only `sysRoles`.

auth/api/grpc/requests.go
```diff
@@ -4,10 +4,14 @@
 package grpc
 
 import (
+	"errors"
+
 	"github.com/MainfluxLabs/mainflux/auth"
 	"github.com/MainfluxLabs/mainflux/internal/apiutil"
 )
 
+var errMissingRole = errors.New("missing role")
```
manuio: Dffine it in the proper place so it can be used by the api  logging middleware

manuio (changes requested):

auth/postgres/init.go
```diff
@@ -104,6 +104,19 @@ func migrateDB(db *sqlx.DB) error {
 					`DROP TABLE IF EXISTS group_relations`,
 				},
 			},
+			{
+				Id: "auth_4",
+				Up: []string{
+					`CREATE TABLE IF NOT EXISTS users_roles (
+					 role VARCHAR(12) CHECK (role IN ('root_admin', 'admin')),
```
manuio: Let's use `root` instead of `root_admin`

manuio (changes requested):

auth/service.go
```diff
@@ -281,7 +273,7 @@ func (svc service) ListOrgs(ctx context.Context, token string, admin bool, pm Pa
 	}
 
 	if admin {
-		if err := svc.Authorize(ctx, AuthzReq{Email: user.Email}); err == nil {
+		if err := svc.Authorize(ctx, AuthzReq{Subject: rootSubject, Token: token, Action: "1"}); err == nil {
```
manuio: Why `"1"`? Don't set any value

manuio (changes requested):

auth/api/grpc/endpoint.go
```diff
@@ -62,10 +62,35 @@ func authorizeEndpoint(svc auth.Service) endpoint.Endpoint {
...
+
+func assignRoleEndpoint(svc auth.Service) endpoint.Endpoint {
+	return func(ctx context.Context, request interface{}) (interface{}, error) {
+		req := request.(assignRoleReq)
+
+		if err := req.validate(); err != nil {
+			return emptyRes{}, err
+		}
+
+		err := svc.AssignRole(ctx, req.ID, req.Role)
```
manuio: Inline the error

auth/service.go
```diff
@@ -783,6 +775,31 @@ func (svc service) Restore(ctx context.Context, token string, backup Backup) err
...
 
+func (svc service) AssignRole(ctx context.Context, id, role string) error {
+	if err := svc.roles.SaveRole(ctx, id, role); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (svc service) isAdmin(ctx context.Context, id string) error {
```
manuio: rename this method `canAccessRoot` and remove `rootSubject`

auth/service.go
```diff
@@ -145,37 +153,23 @@ func (svc service) RetrieveKey(ctx context.Context, token, id string) (Key, erro
...
-		return Identity{ID: key.IssuerID, Email: key.Subject}, nil
-	case APIKey:
-		_, err := svc.keys.Retrieve(context.TODO(), key.IssuerID, key.ID)
-		if err != nil {
-			return Identity{}, errors.ErrAuthentication
-		}
-		return Identity{ID: key.IssuerID, Email: key.Subject}, nil
+	switch ar.Subject {
+	case rootSubject:
+		return svc.rootSubject(ctx, user.ID)
```
manuio: Here use `canAccessRoot`

auth/service.go
```diff
@@ -281,7 +275,7 @@ func (svc service) ListOrgs(ctx context.Context, token string, admin bool, pm Pa
 	}
 
 	if admin {
-		if err := svc.Authorize(ctx, AuthzReq{Email: user.Email}); err == nil {
+		if err := svc.Authorize(ctx, AuthzReq{Subject: rootSubject, Token: token, Action: ""}); err == nil {
```
manuio: You can probably also use `canAccessRoot` here

users/service.go
```diff
@@ -259,12 +305,12 @@ func (svc usersService) ViewProfile(ctx context.Context, token string) (User, er
 }
 
 func (svc usersService) ListUsers(ctx context.Context, token string, pm PageMetadata) (UserPage, error) {
-	user, err := svc.identify(ctx, token)
+	_, err := svc.identify(ctx, token)
 	if err != nil {
 		return UserPage{}, err
 	}
```
manuio: Since ListUsers is only a root action you can remove this identify check, it's useless

users/service.go
```diff
@@ -476,9 +522,9 @@ func (svc usersService) identify(ctx context.Context, token string) (userIdentit
 	return userIdentity{identity.Id, identity.Email}, nil
 }
 
-func (svc usersService) authorize(ctx context.Context, email string) error {
+func (svc usersService) authorize(ctx context.Context, token string) error {
 	req := &mainflux.AuthorizeReq{
```
manuio, line 522: Is this going to work without Subject?

manuio (changes requested):

manuio (approved):
LGTM

PR #243, opened 2023-07-13
MF-212 - Create helper DB methods in auth, things and users.

### LINKED ISSUE #212: Reuse helper methods in auth, things and users
Currently there are helper methods, which are the same, and they are used in `auth`, `things,` and `user` services.
Check them and make them to be reusable.

zzokki81:
Created helper DB methods  which is used in `users`, `things` and `auth` db layer.

Resolves #212


manuio (approved):
LGTM

PR #245, opened 2023-07-14
MF-244 - Change the apiutil logger to log if contains errors

### LINKED ISSUE #244: Change the apiutil logger to log if contains errors
The current system lacks proper logging when errors are wrapped. Instead of solely checking for the occurrence of an error, it is essential to examine the content of the error message itself.

zzokki81:
Added to `apiutil` logger to log if contains errors.

Resolves #244


manuio (approved):
LGTM

PR #256, opened 2023-07-28
MF-255 - Make group members UNIQUE

### LINKED ISSUE #255: Make group members UNIQUE
Change the table field to UNIQUE here: https://github.com/MainfluxLabs/mainflux/blob/master/things/postgres/init.go#L116
Change the method `RetrieveMemberships` to return the groupID where the thing is assigned here: https://github.com/MainfluxLabs/mainflux/blob/master/things/postgres/groups.go#L294 Rename the method to `RetrieveMembership`.

zzokki81:
Added thing per group memberships. Added single thing for `assign` and `unassign`

Resolves #255


manuio (approved):
@zzokki81 Fix the conflict

things/mocks/groups.go
```diff
@@ -34,7 +31,7 @@ type groupRepositoryMock struct {
 func NewGroupRepository() things.GroupRepository {
 	return &groupRepositoryMock{
 		groups:      make(map[string]things.Group),
-		memberships: make(map[string]map[string]things.Group),
+		memberships: make(map[string]string),
 		members:     make(map[string]map[string]map[string]string),
```
manuio: It would be better to use a slice here:`members:     make(map[string][]string)`

manuio (changes requested):

things/mocks/groups.go
```diff
@@ -19,11 +19,8 @@ type groupRepositoryMock struct {
 	// Map of groups, group id as a key.
 	// groups      map[GroupID]auth.Group
 	groups map[string]things.Group
-	// Map of groups (with group id as key) which
-	// represent memberships is element in
-	// memberships' map where member id is a key.
-	// memberships map[MemberID]map[GroupID]auth.Group
-	memberships map[string]map[string]things.Group
+	// Map of group memberships where member id is a key and group id is a value.
+	memberships map[string]string
```
manuio: Rename to `membership`

manuio (changes requested):

manuio (approved):
LGTM

PR #260, opened 2023-08-01
MF-257 - Add unassigned query parameter in group members list

### LINKED ISSUE #257: Add unassigned query parameter in group members list
currently we can list members of a group but not available members to be assigned in the group.
Add a query parameter `unassigned` that will filter things already assigned and return only the things available to be assigned: https://github.com/MainfluxLabs/mainflux/blob/master/things/api/things/http/transport.go#L223

zzokki81:
Added unassigned query parameter in group members list. If it is `true` endpoint will return all unassigned members

Resolves #257


manuio (approved):
LGTM

PR #261, opened 2023-08-01
NOISSUE- Return things instead of members in list members endpoint

zzokki81:
Returned things instead of members in list members endpoint.


manuio (approved):
LGTM

PR #262, opened 2023-08-02
MF- 249 - Fix auth user mock

### LINKED ISSUE #249: Fix users  auth mock
Users auth mock are not well written. There is unused field  `authz`. Also rename user mock file to `authn.go` --> `auth.go`.

### LINKED ISSUE #250: Remove reduntant auth mock file
Currently there is two `auth` mock files which is used for same thing.  Remove the redundant one  and make all services use the same file

zzokki81:
Fix auth user mock

Resolves #249


manuio (approved):
LGTM

PR #265, opened 2023-08-02
MF-264 - Check is user group owner for update and delete group

### LINKED ISSUE #264: Check is user group owner for remove and update group.
Anyone with a valid token can remove and update a group. First, get the group with the given id, then check if the user is the owner of the group. If the user is the owner, the group can be deleted or updated

zzokki81:
Checked is user group owner for update and delete group. If the user is the owner, the group can be deleted or updated.

Resolves #264


manuio (approved):
LGTM

PR #266, opened 2023-08-02
MF-263 - Unassign members before delete group

### LINKED ISSUE #263: Unassign members before delete group
Currently, if we try to delete a group to which a member is assigned, we will get the `internal server error` error. Unassign all members from group before delete it.

zzokki81:
Unassigned all members before delete group.

Resolves #263


manuio (approved):
LGTM

PR #267, opened 2023-08-07
MF-250 - Remove redundant auth mock file

### LINKED ISSUE #250: Remove reduntant auth mock file

zzokki81:
Removed redundant `auth` mock file. Moved `auth.go`  from users to  pkg mocks directory.
Resolves #250


manuio (approved):
LGTM

PR #273, opened 2023-08-09
MF-259 - Add view admin user tests

### LINKED ISSUE #259: Add view admin user tests

zzokki81:
Add tests to see if admin user have correct `root` role.

Resolves #259

### DIFF
--- a/users/service_test.go
+++ b/users/service_test.go
@@ -27,11 +27,11 @@ const (
 var (
 	admin           = users.User{Email: "admin@example.com", ID: "574106f7-030e-4881-8ab0-151195c29f94", Role: "root"}
 	unauthUser      = users.User{Email: "unauthUser@example.com", ID: "6a32810a-4451-4ae8-bf7f-4b1752856eef", Password: "password"}
-	selfRegister    = users.User{Email: "selfRegister@example.com", Password: "password"}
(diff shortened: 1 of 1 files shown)


users/service_test.go
```diff
@@ -204,7 +214,8 @@ func TestViewProfile(t *testing.T) {
 	}
 
 	for desc, tc := range cases {
-		_, err := svc.ViewProfile(context.Background(), tc.token)
+		p, err := svc.ViewProfile(context.Background(), tc.token)
+		assert.Equal(t, tc.role, p.Role, fmt.Sprintf("%s: expected %s got %s\n", desc, tc.role, p.Role))
```
manuio: This is ok but better compare the User, not only the role

users/service_test.go
```diff
@@ -204,7 +214,8 @@ func TestViewProfile(t *testing.T) {
 	}
 
 	for desc, tc := range cases {
-		_, err := svc.ViewProfile(context.Background(), tc.token)
+		p, err := svc.ViewProfile(context.Background(), tc.token)
```
manuio: `u` or `user` instead of `p`

manuio (changes requested):

manuio (approved):
LGTM

Issue #270, opened 2023-08-08
Add Channels to groups


manuio:
Currently we can connect Things to any Channel. It wasn't a problem before orgs and groups but now it's a security issue because it allows a user to connect 2 Things assigned in 2 different groups from 2 different orgs in a same Channel.

To fix it we will introduce Channels in groups and we will verify before any connection that the Thing and the Channel are assigned in the same group.

PR #275, opened 2023-08-24
MF-270 - Move group_channels to new init

zzokki81:
Moved `group_channels` table to new init

This is related to issue:
#270


manuio (approved):
LGTM

PR #290, opened 2023-09-19
MF-289 - Return auth err for connection if thing and channel aren't in the same group

### LINKED ISSUE #289: Return an auth error when thing and channel try to connect if they are not in the same group

zzokki81:
Resolves #289

### DIFF
--- a/things/service.go
+++ b/things/service.go
@@ -495,33 +495,22 @@ func (ts *thingsService) Connect(ctx context.Context, token, chID string, thIDs 
 		return err
 	}
 
-	if cgrID != "" {
-		var tgrIDs []string
-		for _, thID := range thIDs {
-			tgrID, err := ts.groups.RetrieveThingMembership(ctx, thID)
-			if err != nil {
-				return err
-			}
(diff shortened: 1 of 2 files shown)


things/service.go
```diff
@@ -495,31 +495,33 @@ func (ts *thingsService) Connect(ctx context.Context, token, chID string, thIDs
...
+			if err != nil {
+				return err
+			}
 
-	var tgrIDs []string
-	for _, thID := range thIDs {
-		tgrID, err := ts.groups.RetrieveThingMembership(ctx, thID)
-		if err != nil {
-			return err
+			if tgrID != "" {
```
manuio: Check here if groupID is different. No need to create an array for that. As sooner you catch the error as better it is.

manuio (changes requested):

manuio (changes requested):
It's not related to this PR but can you check why `IsChannelOwner` method is exported?

manuio (approved):
LGTM

PR #291, opened 2023-09-20
MF-287 - List disconnected things by channel within the same group

### LINKED ISSUE #287: List things by channels within the same group
System currently  allows users to list all connected and not connected things by channel, without considering their group associations.
Allow things to be listed by channel only if they are members of the same group as the selected channel.

zzokki81:
Added endpoint to list all  disconnected things assigned to the same group as channel.

Resolves #287


manuio (approved):
LGTM

PR #311, opened 2023-09-29
MF-310 - Fix save grup policies

### LINKED ISSUE #310: Fix save grup policies
When a group array tries to be assigned to an organization, it fails because of group policy saving.

zzokki81:
Resolves #310


manuio (approved):
LGTM

PR #313, opened 2023-09-29
MF-312 - Update admin ID and assign root role

### LINKED ISSUE #312: Update admin ID
If restore from backup need to be done admin ID must be updated to match id from backup file. 
If initial admin ID is not updated, admin data from backup file will be not represented.

Change `UpdateUser` method so it can  be used to update admin ID.

zzokki81:
Update admin ID and assign root role

Resolves: #312

### DIFF
--- a/users/postgres/users.go
+++ b/users/postgres/users.go
@@ -37,7 +37,7 @@ func (ur userRepository) Save(ctx context.Context, user users.User) (string, err
 		return "", errors.ErrMalformedEntity
 	}
 
-	dbu, err := toDBUser(user)
+	dbu, err := toDBUser(user.ID, user)
 	if err != nil {
 		return "", errors.Wrap(errors.ErrCreateEntity, err)
 	}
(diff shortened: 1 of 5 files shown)


users/postgres/users.go
```diff
@@ -81,10 +81,15 @@ func (ur userRepository) Update(ctx context.Context, user users.User) error {
 	return nil
 }
 
-func (ur userRepository) UpdateUser(ctx context.Context, user users.User) error {
-	q := `UPDATE users SET id = :id, metadata = :metadata WHERE email = :email AND status = 'enabled'`
+func (ur userRepository) UpdateUser(ctx context.Context, id string, user users.User) error {
+	var adminID string
```
manuio: This is not an ID, it's a part of the SQL query. Better call it `idq`

manuio (changes requested):

users/postgres/users.go
```diff
@@ -81,10 +81,15 @@ func (ur userRepository) Update(ctx context.Context, user users.User) error {
 	return nil
 }
 
-func (ur userRepository) UpdateUser(ctx context.Context, user users.User) error {
-	q := `UPDATE users SET id = :id, metadata = :metadata WHERE email = :email AND status = 'enabled'`
+func (ur userRepository) UpdateUser(ctx context.Context, id string, user users.User) error {
```
manuio: user has ID, no need to pass a new argument

manuio (changes requested):

manuio (approved):
LGTM

PR #316, opened 2023-10-02
MF-315 - Use PATCH for unassign things and channels

### LINKED ISSUE #315: Use PATCH for unassign  things and channels
We use a list of IDs to unassign things and channels. For both endpoints,  send a `PATCH` request  instead of using the `DELETE` method.

zzokki81:
Resolves: #315


manuio (approved):
LGTM

PR #322, opened 2023-10-09
MF-321 - Create policy for creator only

### LINKED ISSUE #321: Create policy only for creator
Currently, when assigning a group to an organization, we create a policy for each member. Instead, create a policy only for the creator.
[https://github.com/MainfluxLabs/mainflux/blob/master/auth/service.go#L618](url)

zzokki81:
Created  policy for creator only.

Resolves: #321

### DIFF
--- a/auth/keys.go
+++ b/auth/keys.go
@@ -22,7 +22,7 @@ var (
 )
 
 const (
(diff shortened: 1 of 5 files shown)


manuio (approved):
LGTM

PR #326, opened 2023-10-12
MF-325 - Create view org member endpoint.

### LINKED ISSUE #325: Create view member endpoint.
Create `viewMember` which will show info of org member.

zzokki81:
Created `viewMember` endpoint which will get member info.

Resolves: #325


manuio (approved):
LGTM

PR #327, opened 2023-10-12
MF-323 - Add permission to admin to see group details.

### LINKED ISSUE #323: Add permission to admin to see group details.
Add permission to admin so he can see group details from all users.

zzokki81:
Added permission to admin so he can see group details.

Resolves: #323


manuio (approved):
LGTM

Issue #330, opened 2023-10-13
Create CRUD for policies.


zzokki81:
Create endponts for `create`, `get`, `update` and `delete` policies

PR #333, opened 2023-10-16
MF-330 - Add update policy endpoint

zzokki81:
Added an `updatePolicy` endpoint that can be used to update group member policies.
#330


manuio (approved):
LGTM

PR #347, opened 2023-10-27
MF-324 -Create gRPC endpoint to get admin role.

### LINKED ISSUE #324: Create gRPC endpoint to fetch admin role.
When launching the user service, we automatically create an admin user, with admin role. During every start it will try to insert role for admin user.  Fix it to check if `admin` role already exists.

zzokki81:
Resolves #324


manuio (approved):
LGTM

PR #357, opened 2023-11-09
MF-354 - Prevent org owner from changing his role

### LINKED ISSUE #354: Owner can change his org role to another role
When a user creates an Org, he become an owner of that Org. Currently, he can change his ownership role to another role. That shouldn't be possible

zzokki81:
Resolves: #354


manuio (approved):
LGTM

PR #364, opened 2023-11-23
MF-360 - Fix orgs DB tables references

### LINKED ISSUE #360: Fix orgs DB tables references
Currently, within our service, we are performing multiple actions such as unassigning members and groups before removing an organization. However, it would be more efficient to confine the unassigning of members and groups directly in the database.

Modify `member_relations` and  `group_relations` to use `DELETE CASCADE`

zzokki81:
Resolves: #360


manuio (approved):
LGTM

PR #366, opened 2023-12-05
MF-365 - Create backup messages endpoint

### LINKED ISSUE #365: Create backup messages endpoint
For backing up messages, a JSON response is currently used, but it doesn't perform well with a large number of messages. To address this, create a `/backup` endpoint in the reader's service. This endpoint should retrieve all messages and write them into a `CSV` file.

zzokki81:
Resolves #365

### DIFF
--- a/readers/api/endpoint.go
+++ b/readers/api/endpoint.go
@@ -7,8 +7,8 @@ import (
 	"bytes"
 	"context"
 	"encoding/csv"
-	"fmt"
 	"reflect"
+	"strconv"
 
 	auth "github.com/MainfluxLabs/mainflux/auth"
 	"github.com/MainfluxLabs/mainflux/pkg/errors"
@@ -115,7 +115,6 @@ func generateCSV(page readers.MessagesPage) ([]byte, error) {
 	var buf bytes.Buffer
 	writer := csv.NewWriter(&buf)
 
-	// Write header row
 	header := []string{
 		"ID",
 		"Channel",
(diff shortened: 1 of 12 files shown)


readers/api/endpoint.go
```diff
@@ -80,3 +110,118 @@ func restoreEndpoint(svc readers.MessageRepository) endpoint.Endpoint {
...
+	}
+	return defaultValue
+}
+
+func getBoolValue(ptr *bool, defaultValue string) string {
+	if ptr != nil {
+		return fmt.Sprintf("%v", *ptr)
+	}
+	return defaultValue
+}
```
manuio: This can be probably done with a single method:
```
func valueToString(ptr *bool, defaultValue string) string {
	if ptr != nil {
		return fmt.Sprintf("%v", *ptr)
	}
	return defaultValue
}
```

readers/api/endpoint.go
```diff
@@ -61,6 +65,32 @@ func listAllMessagesEndpoint(svc readers.MessageRepository) endpoint.Endpoint {
...
 }
 
+func backupEndpoint(svc readers.MessageRepository) endpoint.Endpoint {
+	return func(ctx context.Context, request interface{}) (interface{}, error) {
+		req := request.(listAllMessagesReq)
+		if err := req.validate(); err != nil {
+			return nil, err
+		}
+
+		// Check if user is authorized to read all messages
```
manuio: No need for this comment. `authorizeAdmin` is self explenatory

manuio (changes requested):

manuio (approved):
LGTM

PR #377, opened 2024-01-15
NOISSUE - Read timefield from profile in Transform

zzokki81:
Read  profile time-field in  json `Trasform` method


manuio (approved):
LGTM

PR #391, opened 2024-02-19
MF-390 - Store message in DB when notify

### LINKED ISSUE #390: Store message in DB when notify
Enhance the current notifier implementation to allow simultaneous publishing and notification when `retention` is set to `true`. If `retention` is set to `false`, it should only notify.

zzokki81:
Resolves #390


manuio (approved):
LGTM

PR #404, opened 2024-03-01
MF-403 - Use only profile instead of conn in publish

### LINKED ISSUE #403: Use only profile instead  of conn in publish
Passing whole `conn` in Publish  is redundant. Remove it and use profile instead.

zzokki81:
Resolves: #403


manuio (approved):
LGTM

PR #410, opened 2024-03-17
MF-408 - Create isThingOwner gRPC endpoint

### LINKED ISSUE #408: Create isThingOwner gRPC endpoint
Currently we can validate channels ownership from user token but not things ownership. Create a gRPC endpoint similar to isChannelOwner that validate thing ownership from user token.

majabirmancevic:
Resolves #408


manuio (approved):
LGTM

PR #451, opened 2024-04-26
MF-447 - Fix delete orgs

### LINKED ISSUE #447: DELETE Org
I created an org and when I wanted to delete it I could not and I got the following error message {“error”:“org is not empty”}

edufgg:
Resolves #447


manuio (approved):
LGTM

PR #520, opened 2024-09-03
MF-518 - Add pagination to the webhooks and notifiers services

### LINKED ISSUE #518: Add pagination to the webhooks and notifiers services
There is no page metadata within responses for retrieving a list of webhooks/notifiers. It needs to be added.

majabirmancevic:
Resolves #518


manuio (approved):
LGTM

PR #532, opened 2024-10-01
MF-527 - Fix group rights in Remove service methods

### LINKED ISSUE #527: Fix group rights in Remove service methods
Currently, the groupID is passed as a parameter through a path, and based on it, the rights of the logged-in user in that group are checked. For each entity passed for deletion, the groupID should be checked and whether the logged-in user has the appropriate role for that group.

majabirmancevic:
Resolves #527


manuio (approved):
LGTM

PR #570, opened 2025-01-22
MF-545 - Add cli command for getting metadata by key

majabirmancevic:
Add missing CLI command for getting thing metadata by key: #545


manuio (approved):
LGTM

PR #591, opened 2025-02-25
MF-586 - Split Authorize method

### LINKED ISSUE #586: Split Authorize method
Currently, the `Authorize` method checks the user's access rights for a group and its entities.
Instead, split that method so that there is a separate method for a specific entity, for example:
`CanUserAccessThing`,  `CanUserAccessProfile`, `CanUserAccessGroup`

majabirmancevic:
Resolves #586


manuio (approved):
LGTM

PR #599, opened 2025-02-28
MF-587 - Move database methods to the pkg package

### LINKED ISSUE #587: Move database methods to the pkg package
Instead of adding [these methods](https://github.com/MainfluxLabs/mainflux/blob/5da90df30110a35ee1e4ed51f9881c0ce3efa52c/things/postgres/tracing.go#L18) to each service, move them to the `pkg` package and use them from there.

majabirmancevic:
Resolves #587


manuio (approved):
LGTM

Issue #526, opened 2024-09-17
Move commonly used consts to pkg package


majabirmancevic:
Commonly used constants should be in the pkg package and used from there, such as nameKey, offsetKey, limitKey, and others.

manuio:
Resolved by: https://github.com/MainfluxLabs/mainflux/pull/603

PR #603, opened 2025-03-06
MF-602 - Use a helper method to create PageMetadata from a request

### LINKED ISSUE #602: Use a helper method to create PageMetadata from a request

majabirmancevic:
Resolves #602 and commonly used constants are moved to pkg/apiutil (#526)


manuio (approved):
LGTM

PR #618, opened 2025-03-21
MF-611 - Add missing tests for groups and group members

### LINKED ISSUE #611: Add missing tests for groups and group members
There are no service and endpoint tests for groups and group members.

majabirmancevic:
Resolves: #611


manuio (approved):
LGTM

PR #634, opened 2025-04-25
MF-633 - Remove Transformer interface

### LINKED ISSUE #633: Remove Transformer interface
We currently use the Transformer interface to transform messages from adapters before consumers consume them.
Instead, the payload from the Mainflux Message needs to be transformed before publishing the Message to a specific topic.

So, instead of the Transformer interface, we would only use the methods from the `pkg` package.
Also, the Mainflux Message needs to be mapped to a JSON or SenML message in the Writer service, depending on the Content-Type defined in the Message.

majabirmancevic:
Resolves #633


manuio (approved):
LGTM

PR #635, opened 2025-04-29
MF-617 - Return ErrAuthorization in getGroupIDByThingID if ID is empty

### LINKED ISSUE #617: GetGroupIDByThingID returns ErrNotFound if ID is empty
GetGroupIDByThingID should return ErrAuthorization

kimanimichael:
### What does this do?
Ensures that GetGroupIDByThingID returns ErrAuthorization if ID is empty

Resolves #617 

No

No

### DIFF
--- a/things/service.go
+++ b/things/service.go
@@ -817,6 +817,9 @@ func (ts *thingsService) canAccessOrg(ctx context.Context, token, orgID, subject
 }
 
 func (ts *thingsService) getGroupIDByThingID(ctx context.Context, thID string) (string, error) {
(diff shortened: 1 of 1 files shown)


manuio:
@kimanimichael Thak you for your PR but actually the description of the issue was empty and your fix isn't really good. If I remember well the issue was in the gRPC server. I should have described this better... 
@majabirmancevic What you think about? Should we close this one or should we propose another fix here?

majabirmancevic:
Let's close and investigate this issue by expanding its description.

PR #636, opened 2025-04-30
NOISSUE - Update downlink section in docker-compose

majabirmancevic:
Add params for rules gRPC client in the downlink section.


manuio (approved):
LGTM

PR #646, opened 2025-05-30
MF-645 - Return NotFound when ID doesn't exist for orgs and groups

### LINKED ISSUE #645: Return NotFound when ID doesn't exist for orgs and groups
Currently, if an invalid UUID is passed to get an org or group by ID, a 500 error is returned.
Extend the check in RetrieveByID to return a 404 error.

majabirmancevic:
Resolves #645


manuio (approved):
LGTM

PR #658, opened 2025-06-10
MF-657 - CLI: Explicitly set URLs for every service when initializing the SDK

### LINKED ISSUE #657: Unable to create Groups using the CLI
Currently, attempting to create a Group using the CLI tool, e.g.:

```sh
mainfluxlabs-cli groups create \ 
    '{"name":"<group_name>","description":"<group_description>","metadata":{}}' \
     <ORG_ID> <admin_access_token>
```

Where `<ORG_ID>` is an ID of a valid, existing Org and `<admin_access_token>` is a valid access token for a root admin account, results in the following output:

```
error: failed to create entity : 404 Not Found
```

This appears to be because the CLI is initializing the SDK with the config field `thingsURL` set to a [hardcoded `http://localhost`](https://github.com/MainfluxLabs/mainflux/blob/4f67f8a55acfd3a3bf1b1fafb5012931a57070da/cmd/cli/main.go#L21), with no prefix for the Things service.

A temporary workaround is to manually set the URL of the Things service when running the CLI using the `--things-service`/`-t` flag. For example:

```
mainflux-cli groups --things-url <PROPER_URL> ...
```

However, we should probably update the default URLs of the CLI to match those of other services.

bool3max:
This PR fixes #657 and other possible CLI issues by explicitly initializing the URL for each service in the SDK config.

### DIFF
--- a/cmd/cli/main.go
+++ b/cmd/cli/main.go
@@ -17,11 +17,11 @@ const defURL string = "http://localhost"
 func main() {
 	msgContentType := string(sdk.CTJSONSenML)
 	sdkConf := sdk.Config{
-		AuthURL:         defURL,
-		ThingsURL:       defURL,
+		AuthURL:         fmt.Sprintf("%s/svcauth", defURL),
+		ThingsURL:       fmt.Sprintf("%s/svcthings", defURL),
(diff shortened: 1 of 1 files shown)


manuio (approved):
LGTM

PR #667, opened 2025-06-12
MF-659 - Validate user password in User.Validate

### LINKED ISSUE #659: User.Validate() method doesn't validate password against regex
The [`users.User.Validate()`](https://github.com/MainfluxLabs/mainflux/blob/164fbfcde8aabc78ec4487cb88f1c974e880ef90/users/users.go#L29) method currently validates the e-mail, and checks that the password is non-zero length, but **doesn't match it against the password regex**. 

Note: Checking the password against the regex when registering new Users seems to currently be done manually [at the `users` service level](https://github.com/MainfluxLabs/mainflux/blob/164fbfcde8aabc78ec4487cb88f1c974e880ef90/users/service.go#L229), *as well as* at [the HTTP endpoint level](https://github.com/MainfluxLabs/mainflux/blob/164fbfcde8aabc78ec4487cb88f1c974e880ef90/users/api/http/endpoint.go#L35). However, as the `Validate()` method is used in tests as well, it would be a good idea to update it to properly validate the password as well.

bool3max:
This PR currently:

- ~~Sets and exports the default user password validation regex string in one place~~
    - ~~Previously the same constant was repeated in about 4 different places~~
- Validates the user's password against the regex in `users.User.Validate()`
- ~~Adds a couple password-related tests to the `users` package~~

---

However, it is currently broken as `users.User.Validate()` doesn't accept the regex as an argument, but instead always checks against the default one (this is problematic because users can set the regex at runtime using the `MF_USERS_PASS_REGEX` env. variable). Making it accept a regex is trivial, but as `User.Validate()` is currently called [when HTTP request data is validated in the handlers](https://github.com/MainfluxLabs/mainflux/blob/309db4fa142607721132ae8573306639fac6d670/users/api/http/requests.go#L30), and as those methods have no way to inspect the password regex being used at runtime, some refactoring would need to be done. 

There are a couple options:

- Remove calls to `Users.Validate()` from the HTTP handler layer ([requests.go](https://github.com/MainfluxLabs/mainflux/blob/master/users/api/http/requests.go)) and move them to the users service layer methods (`users.Service.Register()`, `users.Service.SelfRegister()`, etc...)
    - Methods in the users service implementation already perform password checks
    - Would require some work to match the current behavior mandated by endpoint and service tests
    - Would leave some struct `.validate` methods in [requests.go](https://github.com/MainfluxLabs/mainflux/blob/master/users/api/http/requests.go) empty - kinda weird but maybe OK?
- Keep `Users.Validate()` calls in the HTTP handler layer (in [requests.go](https://github.com/MainfluxLabs/mainflux/blob/master/users/api/http/requests.go), as it currently is), but refactor the structure such that they are able to be passed the password regex from the `users.Service` implementation
    - Probably unnecessarily complicated
- Abandon this entirely, keep password checks in users service methods without the use of `User.Validate()`
    - This leaves the `User.Validate()` somewhat pointless as it only checks for validity of the e-mail

### DIFF
--- a/cmd/users/main.go
+++ b/cmd/users/main.go
@@ -165,7 +165,7 @@ func main() {
 	svc := newService(db, dbTracer, auth, cfg, logger)
 
 	g.Go(func() error {
-		return servershttp.Start(ctx, httpapi.MakeHandler(svc, usersHttpTracer, logger), cfg.httpConfig, logger)
+		return servershttp.Start(ctx, httpapi.MakeHandler(svc, usersHttpTracer, logger, cfg.passRegex), cfg.httpConfig, logger)
 	})
 
 	g.Go(func() error {
--- a/cmd/users/main.go
+++ b/cmd/users/main.go
@@ -290,7 +290,7 @@ func newService(db *sqlx.DB, tracer opentracing.Tracer, ac protomfx.AuthServiceC
 
 	idProvider := uuid.New()
 
-	svc := users.New(userRepo, hasher, ac, emailer, idProvider, c.passRegex)
+	svc := users.New(userRepo, hasher, ac, emailer, idProvider)
 	svc = httpapi.LoggingMiddleware(svc, logger)
 	svc = httpapi.MetricsMiddleware(
 		svc,
--- a/pkg/sdk/go/users_test.go
+++ b/pkg/sdk/go/users_test.go
@@ -49,7 +49,7 @@ func newUserService() users.Service {
 	auth := mocks.NewAuthService(admin.ID, usersList, orgsList)
 	emailer := usmocks.NewEmailer()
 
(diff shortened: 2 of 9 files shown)


bool3max:
Eventually we landed on a variation of the second option:

The responsibility of both e-mail and password validity checks is now done only in the HTTP handlers (using the `requests` structs and their [`.validate()`](https://github.com/bool3max/mainflux/blob/a4369758cec2b87141f0bd17201ff2f47123bedc/users/api/http/requests.go#L21) methods which call `User.Validate()`). The `users.Service` methods that deal with passwords and e-mails (e.g. `.SelfRegister()`, `.Register()`) now assume that what they receive is fully valid.

manuio (approved):
LGTM

PR #673, opened 2025-06-16
MF-671 - Change error on invalid old password when changing password

### LINKED ISSUE #671: Change password returns 401 Unauthorized on wrong old password

anna02272:
Resolves #671


pkg/errors/types.go
```diff
@@ -39,4 +39,7 @@ var (
 
 	// ErrMessage indicates an error converting a message to Mainflux message.
 	ErrMessage = New("failed to convert to Mainflux message")
+
+	// ErrInvalidPassword indicates that current password is invalid.
+	ErrInvalidPassword = New("invalid current password")
```
manuio, line 44: Use this error instead of the one defined in apiutil. apiutil errors should be used onlyin api packages

manuio (changes requested):

users/service.go
```diff
@@ -441,7 +441,7 @@ func (svc usersService) ChangePassword(ctx context.Context, token, email, passwo
 	// Admin changes password for another user
 	case oldPassword == "" && email != "":
 		if err := svc.isAdmin(ctx, token); err != nil {
-			return errors.ErrAuthentication
+			return errors.ErrAuthorization
```
manuio: Could we use the error returned by isAdmin here?

manuio (changes requested):

manuio (approved):
LGTM

PR #677, opened 2025-06-17
MF-674 - Standardize SDK method signatures

### LINKED ISSUE #674: Standardize SDK method names and argument orders
Existing [SDK methods](https://github.com/MainfluxLabs/mainflux/blob/f780687355b8257cf7b14e51cc6dda709a614190/pkg/sdk/go/sdk.go#L173) have non-standardized names as well as arguments and their orders. For example:

- In methods that accept both a `token` argument and page metadata arguments sometimes the token comes first, and sometimes the page metadata
- Page metadata arguments are sometimes passed in as `PageMetadata` struct, and sometimes as two individual `offset` and `limit` arguments
- Methods that retrieve a single resource sometimes have a `View` prefix and sometimes not (e.g. `Group`, `ViewGroupByProfile`)

Considering that the SDK is an external package it would be good to standardize these things.

bool3max:
This PR standardizes SDK method names, their arguments and argument orders:

- Methods that *retrieve* a single resource now always begin with a `Get` prefix
    - Example: `GetThing()`
- Methods that *retrieve* multiple resources now always begin with a `List` prefix
    - Example: `ListThings()`
- In general methods now begin with a verb describing the operation (i.e. `Create`, `Update`, `Delete`, etc..), followed by the resource name
- Methods that accept an authentication `token` now always receive it as the last argument
- Methods that accept pagination metadata now always do so via the use of a `PageMetadata` type, that comes as the second-to-last argument (before the above mentioned token)

Resolves #674

### DIFF
--- a/cli/groups.go
+++ b/cli/groups.go
@@ -59,7 +59,7 @@ var cmdGroups = []cobra.Command{
 					Offset: uint64(Offset),
 					Limit:  uint64(Limit),
 				}
-				gp, err := sdk.GetGroups(meta, args[1])
+				gp, err := sdk.ListGroups(meta, args[1])
 				if err != nil {
 					logError(err)
 					return
@@ -78,7 +78,7 @@ var cmdGroups = []cobra.Command{
 					return
 				}
 
-				res, err := sdk.GetGroupsByOrg(
+				res, err := sdk.ListGroupsByOrg(
 					args[1],
 					mfxsdk.PageMetadata{Offset: uint64(Offset), Limit: uint64(Limit)},
 					args[2],
@@ -146,7 +146,7 @@ var cmdGroups = []cobra.Command{
 				Limit:  uint64(Limit),
 			}
 
-			up, err := sdk.GetThingsByGroup(args[0], meta, args[1])
+			up, err := sdk.ListThingsByGroup(args[0], meta, args[1])
 			if err != nil {
 				logError(err)
 				return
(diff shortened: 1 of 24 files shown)


pkg/sdk/go/keys.go
```diff
@@ -83,7 +83,7 @@ func (sdk mfSDK) Revoke(id, token string) error {
 	return nil
 }
 
-func (sdk mfSDK) RetrieveKey(id, token string) (retrieveKeyRes, error) {
+func (sdk mfSDK) RetrieveKey(token, id string) (retrieveKeyRes, error) {
```
manuio: Why token is the first arg here?

manuio (changes requested):

pkg/sdk/go/keys.go
bool3max: Mistake -- just swapped them in latest commit.

cli/groups.go
```diff
@@ -140,7 +140,13 @@ var cmdGroups = []cobra.Command{
...
 				return
 			}
-			up, err := sdk.ListThingsByGroup(args[0], args[1], uint64(Offset), uint64(Limit))
+
+			meta := mfxsdk.PageMetadata{
+				Offset: uint64(Offset),
+				Limit:  uint64(Limit),
+			}
+
+			up, err := sdk.GetThingsByGroup(args[0], meta, args[1])
```
manuio: I'd keep `List` naming for lists. I agree to switch from `View` to `Get` when we are fetching an unique entity but it's more readable to use `List` when fetching several entitites.  Plus, we are already using `List` in almost all services methods so it's more consistent. Otherwise we should rename many methods.

manuio (changes requested):

cli/groups.go
bool3max: Changed in latest commit.

manuio (approved):
LGTM

PR #687, opened 2025-06-19
MF-682 - Filter org and group members by email 

### LINKED ISSUE #682: Add Members list filtering by email

anna02272:
Resolves #682


things/members.go
```diff
@@ -133,7 +133,7 @@ func (ts *thingsService) ListGroupMembers(ctx context.Context, token, groupID st
 	page := GroupMembersPage{
 		GroupMembers: gms,
 		PageMetadata: apiutil.PageMetadata{
-			Total:  gpp.Total,
+			Total:  uint64(len(gms)),
```
manuio: Why are we not using `gpp.Total` anymore?

auth/api/http/members/endpoint_test.go
```diff
@@ -520,6 +520,26 @@ func TestListMembers(t *testing.T) {
...
+					Email: viewerEmail,
+					Role:  auth.Viewer,
+				},
+			},
+		},
+		{
+			desc:   "list org members filtered by email that doesn't match",
+			token:  token,
+			status: http.StatusOK,
+			url:    fmt.Sprintf("%s/orgs/%s/members?email=%s", ts.URL, or.ID, "xyz"),
```
manuio: Use a const for `xyz`

manuio (changes requested):

things/members.go
anna02272: Because the final list includes only members with valid emails after filtering. gpp.Total reflects all members before filtering, so using it here would return the count of all members, not just those included in the response.

manuio: `Total` represents the total of elements in the database after filtering not the total of elements in the response.  `len(gms)` is equal to`gpp.Limit`.

things/members.go
```diff
@@ -117,7 +117,7 @@ func (ts *thingsService) ListGroupMembers(ctx context.Context, token, groupID st
 		for _, gp := range gpp.GroupMembers {
 			email, ok := emails[gp.MemberID]
```
manuio: Is there a better way to map IDs and emails?

auth/members.go
```diff
@@ -235,9 +236,10 @@ func (svc service) ListMembersByOrg(ctx context.Context, token string, orgID str
 	mpg := OrgMembersPage{
 		OrgMembers: oms,
 		PageMetadata: apiutil.PageMetadata{
-			Total:  omp.Total,
+			Total:  uint64(len(page.Users)),
```
manuio: Same remark here, you are replacing Total by the total number of user per page.

manuio (changes requested):

manuio (approved):
LGTM

PR #693, opened 2025-06-24
MF-690 - Sort users and members list by email and id

### LINKED ISSUE #690: Add Users and Members list sorting

anna02272:
Resolves #690


users/api/grpc/server.go
```diff
@@ -61,7 +61,7 @@ func (s *grpcServer) GetUsersByEmails(ctx context.Context, req *protomfx.UsersBy
 
 func decodeGetUsersByIDsRequest(_ context.Context, grpcReq interface{}) (interface{}, error) {
 	req := grpcReq.(*protomfx.UsersByIDsReq)
-	return getUsersByIDsReq{ids: req.GetIds()}, nil
+	return getUsersByIDsReq{ids: req.GetIds(), order: req.Order, dir: req.Dir}, nil
```
manuio, line 64: I understand why you are doing it here but I wonder if this shouldn't be moved to users service and to fetch the IDs of a certain orgID from there. We recently added endpoints to list things, profiles and groups per org in things service. We need the same in users service.

manuio (changes requested):

users/api/grpc/server.go
anna02272: We can move it to the users service, but the users table currently has only id, email, password, metadata, and status - no org reference. Our auth service manages orgs and user memberships via orgs and member_relations tables. To implement ListUsersByOrg, we’d either need to extend the users schema or query org membership from auth ??

manuio (approved):
LGTM

PR #698, opened 2025-06-27
NOISSUE - Document schemas of Postgres databases

bool3max:
This PR creates a `README.md` file in each service's `postgres/` directory, describing that database's schema. 

For simple schemas, i.e. ones with few migrations in the `postgres/init.go` source file, the final schema was mostly copy-pasted based on the migrations.

For more complex ones, i.e. ones with multiple migrations (such as [`things/postgres/init.go`](https://github.com/MainfluxLabs/mainflux/blob/58726211da7ea7933598c5d7d7ac9eb3a9002955/things/postgres/init.go)), an up to date schema was created using the `pg_dump` utility on a running db container. Unfortunately, while `pg_dump` produces a valid and reproducible schema, it's not as readable as human-written SQL - e.g. table constraints are added one by one using `ALTER TABLE ... ADD CONSTRAINT` statements instead of right inside the associated `CREATE TABLE` statement, so these schemas could be manually cleaned up to be more readable.

### DIFF
--- a/auth/postgres/README.md
+++ b/auth/postgres/README.md
@@ -1,0 +1,55 @@
+# Auth - DB Schema
+
+```sql
+CREATE TABLE public.keys (
+    id character varying(254) NOT NULL,
+    type smallint,
+    subject character varying(254) NOT NULL,
+    issuer_id uuid NOT NULL,
+    issued_at timestamp without time zone NOT NULL,
+    expires_at timestamp without time zone
+);
+
+CREATE TABLE public.member_relations (
+    member_id uuid NOT NULL,
+    org_id uuid NOT NULL,
+    role character varying(10) NOT NULL,
+    created_at timestamp with time zone,
+    updated_at timestamp with time zone
+);
+
+CREATE TABLE public.orgs (
+    id uuid NOT NULL,
+    owner_id uuid NOT NULL,
+    name character varying(254) NOT NULL,
+    description character varying(1024),
+    metadata jsonb,
+    created_at timestamp with time zone,
+    updated_at timestamp with time zone
+);
+
+CREATE TABLE public.users_roles (
+    role character varying(12),
+    user_id uuid NOT NULL,
+    CONSTRAINT users_roles_role_check CHECK (((role)::text = ANY ((ARRAY['root'::character varying, 'admin'::character varying])::text[])))
+);
+
+ALTER TABLE ONLY public.keys
+    ADD CONSTRAINT keys_pkey PRIMARY KEY (id, issuer_id);
+
+ALTER TABLE ONLY public.member_relations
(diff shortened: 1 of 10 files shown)


auth/postgres/README.md
```diff
@@ -0,0 +1,55 @@
+# Auth - DB Schema
+
+```sql
+CREATE TABLE public.keys (
```
manuio: Why are we using `public` prefix?

manuio (changes requested):

auth/postgres/README.md
bool3max: It was generated by 'pg_dump', but I can remove it.

readers/postgres/README.md
```diff
@@ -2,6 +2,33 @@
 
 Postgres reader provides message repository implementation for Postgres.
 
+## Readers - DB Schema
+
+```sql
+CREATE TABLE public."json" (
```
manuio: Why `"json"` has double quote but not `messages`?

readers/postgres/README.md
```diff
@@ -2,6 +2,33 @@
...
+    publisher uuid NOT NULL,
+    protocol text,
+    name text NOT NULL,
+    unit text,
+    value double precision,
+    string_value text,
+    bool_value boolean,
+    data_value text,
+    sum double precision,
+    "time" double precision NOT NULL,
```
manuio: Why time has double quote?

auth/postgres/README.md
```diff
@@ -0,0 +1,55 @@
...
+    updated_at timestamp with time zone
+);
+
+CREATE TABLE public.users_roles (
+    role character varying(12),
+    user_id uuid NOT NULL,
+    CONSTRAINT users_roles_role_check CHECK (((role)::text = ANY ((ARRAY['root'::character varying, 'admin'::character varying])::text[])))
+);
+
+ALTER TABLE ONLY public.keys
```
manuio: Why we are documenting this? We only need the current DB init.

manuio (changes requested):

readers/postgres/README.md
bool3max: Also an artifact of `pg_dump`, probably because `json` is a reserved keyword of some kind in Postgres. I will unquote it.

readers/postgres/README.md
bool3max: Same reason as above, but I will unquote it here in the docs.

auth/postgres/README.md
bool3max: This is because the human-written column in the migrations `init.go` file has a column defined like so:

 `role    VARCHAR(12) CHECK (role IN ('root', 'admin'))`

But the above output is generated by `pg_dump` and is therefore uglier. I'll rewrite it.

manuio (approved):
LGTM

PR #702, opened 2025-07-01
MF-699 - Implement provisioning Things and Profiles from CSV

### LINKED ISSUE #699: CLI: Fully implement provisioning Things and Profiles from CSV files
Provisioning Things and Profiles using the CLI's `provision` subcommand is currently not fully implemented: in both cases the function for parsing the CSV file only ever reads the first column of each entry.

We need to define and document the column order for both Things and Profiles and finish the implementation.

bool3max:
This PR finishes the implementation of provisioning functionality when using CSV input files.

### DIFF
--- a/cli/provision.go
+++ b/cli/provision.go
@@ -26,28 +26,28 @@ const csvThingsFieldCount = 4
 
 // These constants define the order of the CSV columns (fields) of records containing Things to be provisioned
 const (
-	csvThingsFieldID = iota
-	csvThingsFieldName
-	csvThingsFieldGroupID
-	csvThingsFieldProfileID
+	thingID = iota
+	thingName
+	thingGroupID
+	thingProfileID
 )
 
 const csvProfilesFieldCount = 12
 
 // These constants define the order of the CSV columns (fields) of records containing Profiles to be provisioned
 const (
-	csvProfilesFieldID = iota
-	csvProfilesFieldName
-	csvProfilesFieldGroupID
-	csvProfilesFieldConfigContentType
-	csvProfilesFieldConfigWrite
-	csvProfilesFieldConfigWebhook
-	csvProfilesFieldConfigTransformerDataFilters
-	csvProfilesFieldConfigTransformerDataField
-	csvProfilesFieldConfigTransformerTimeField
-	csvProfilesFieldConfigTransformerTimeFormat
-	csvProfilesFieldConfigTransformerTimeLocation
-	csvProfilesFieldConfigSMTPID
+	profileID = iota
+	profileName
+	profileGroupID
+	profileConfContentType
+	profileConfWrite
+	profileConfWebhook
+	profileConfTransformerDataFilters
+	profileConfTransformerDataField
+	profileConfTransformerTimeField
+	profileConfTransformerTimeFormat
+	profileConfTransformerTimeLocation
+	profileConfSMTPID
 )
 
 var cmdProvision = []cobra.Command{
@@ -249,10 +249,10 @@ func thingsFromFile(path string) ([]mfxsdk.Thing, error) {
 			}
 
 			thing := mfxsdk.Thing{
-				Name:      record[csvThingsFieldName],
-				ID:        record[csvThingsFieldID],
-				ProfileID: record[csvThingsFieldProfileID],
-				GroupID:   record[csvThingsFieldGroupID],
+				Name:      record[thingName],
+				ID:        record[thingID],
+				ProfileID: record[thingProfileID],
+				GroupID:   record[thingGroupID],
 			}
 
 			recordMetadata := record[csvThingsFieldCount:]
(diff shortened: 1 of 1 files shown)


bool3max:
Provisioning _basic information_ about Things and Profiles from CSV files has been implemented.

As an example, below are the contents of a CSV file that provisions the system with three Things:

```csv
edfd205b-70ed-49e0-a9d3-5f548d10efc0,prTestThing0,90388780-e879-4d40-8ed5-fd859646f310,2995e81b-5da9-4c90-8ee4-99d78ed7478d
,prTestThing1,90388780-e879-4d40-8ed5-fd859646f310,0e865ef2-d5e5-4228-93c9-85c338c683fc,country,Poland,city,Warsaw
,prTestThing2,90388780-e879-4d40-8ed5-fd859646f310,0e865ef2-d5e5-4228-93c9-85c338c683fc,lat,40.7128,lon,70.0612
``` 

An arbitrary number of `metadata` key-value pairs can be specified after the essential fields of each row / record. E.g. the second Thing's metadata object would be: `{"country": "Poland", "city": "Warsaw"}`

---

What's left is to implement specifying [Profiles' `config`](https://mainfluxlabs.github.io/docs/messaging/#configure-profile-config) field in individual CSV rows / records. This is a bit awkward considering that the `config` object within itself contains the `transformer.data_filters` array which can contain an arbitrary number of elements.

We could solve this by designating each field from the `config` object its own field in the CSV (see table below), and define in advance a custom separator for each member of the `data_filters` array (for example the `#` character, see below).

##### Profile provision CSV field order
| ID | Name | Group ID | ... | Config - Write | Config - Transformer - data_filters | ... |
|----|------|----------|-----|----------------|-------------------------------------|-----|
|    |      |          |     |                | val1#val2#val3#val4                 |     |

In the above example, the `val1#val2#val3#val4` value in the appropriate CSV field would result in the following `config` object:

```json
{
  "config": {
    ...
    "transformer": {
      "data_filters": ["val1", "val2", "val3", "val4"],
      ...
    },
    ...
  }
}

```

Or, we could perhaps skip implementing specifications of a Profile's `config` definition in CSV files as it's maybe too complicated/cluttered for the format? @manuio - what are your thoughts on this?

manuio:
@bool3max We can define the Confgi as a JSON to avoid having to define field by field. It's maybe less readable but it would probably simplify the parsing.

bool3max:
In the end we settled on assigning each Profile `config` field its own CSV field, and fields that support multiple values (such as `data_filters`) use the comma character (`,`) as the separator, inside a "" quoted CSV field.

cli/provision.go
```diff
@@ -20,13 +22,41 @@ import (
...
+	csvProfilesFieldGroupID
+	csvProfilesFieldConfigContentType
+	csvProfilesFieldConfigWrite
+	csvProfilesFieldConfigWebhook
+	csvProfilesFieldConfigTransformerDataFilters
+	csvProfilesFieldConfigTransformerDataField
+	csvProfilesFieldConfigTransformerTimeField
+	csvProfilesFieldConfigTransformerTimeFormat
+	csvProfilesFieldConfigTransformerTimeLocation
+	csvProfilesFieldConfigSMTPID
```
manuio: This names are too long. You can probably shorter them by using singular, removing `csv` prefix and `Field`. 
For example, `profileConfTransformerDataFilters`.

manuio (changes requested):

cli/provision.go
```diff
@@ -250,22 +301,69 @@ func profilesFromFile(path string) ([]mfxsdk.Profile, error) {
...
+			}
+			profile.Config["write"] = writeBool
+
+			webhookBool, err := strconv.ParseBool(record[profileConfWebhook])
+			if err != nil {
+				return []mfxsdk.Profile{}, err
+			}
+			profile.Config["webhook"] = webhookBool
+
+			profile.Config["smtp_id"] = record[profileConfSMTPID]
```
majabirmancevic: The `webhook` and `smtp_id` fields are no longer needed in the `config`

bool3max: > The `webhook` and `smtp_id` fields are no longer needed in the `config`

When have they been removed? We should probably update the docs to reflect that change. Either way I'll remove them from the CSV structure

majabirmancevic: It was removed some time ago, but obviously the docs haven't been updated..

cli/provision.go
```diff
@@ -20,13 +22,41 @@ import (
...
+
+// These constants define the order of the CSV columns (fields) of records containing Profiles to be provisioned
+const (
+	profileID = iota
+	profileName
+	profileGroupID
+	profileConfContentType
+	profileConfWrite
+	profileConfWebhook
+	profileConfTransformerDataFilters
```
majabirmancevic, line 45: Consts could be even shorter:
`profileGroup` (without ID)
`confContentType` (without profile)
`confTransformerDataFilters`
etc etc

majabirmancevic: Consts could be even shorter:
`profileGroup` (without ID),
`confContentType` (without profile),
`confTransformerDataFilters`, etc etc

manuio (approved):
LGTM

PR #724, opened 2025-07-07
 MF-704 - Add search endpoints in all services

### LINKED ISSUE #704: Add search endpoints in all services
Currently we can only search complex metadata in things service but it's missing for other entities. Add it where it's missing. 

Here is the things/search endpoint example: https://github.com/MainfluxLabs/mainflux/blob/master/things/api/http/things/transport.go#L80

Nesa002:
Fixed old search things method to validate input and give default values. Added search endpoints for things, profiles and groups. Resolves #704


pkg/apiutil/transport.go
```diff
@@ -304,6 +304,73 @@ func BuildPageMetadata(r *http.Request) (PageMetadata, error) {
...
+		Limit:  DefLimit,
+		Order:  IDOrder,
+		Dir:    DescDir,
+	}
+
+	if r.Body == nil || r.ContentLength == 0 {
+		return pm, nil
+	}
+
+	var body struct {
```
manuio: Use this structure instead of defining an anonimous one: https://github.com/MainfluxLabs/mainflux/blob/master/pkg/apiutil/transport.go#L41

things/api/http/groups/transport.go
```diff
@@ -164,6 +178,34 @@ func decodeListByOrg(_ context.Context, r *http.Request) (interface{}, error) {
...
 
+func decodeListByMetadata(_ context.Context, r *http.Request) (interface{}, error) {
+	pm, err := apiutil.BuildPageMetadataFromBody(r)
+	if err != nil {
+		return nil, err
+	}
+
+	req := listReq{
+		token:        apiutil.ExtractBearerToken(r),
+		pageMetadata: pm}
```
manuio: Close the structure after a new line:
```
req := listReq{
		token:        apiutil.ExtractBearerToken(r),
		pageMetadata: pm,
	}
```

things/api/http/profiles/transport.go
```diff
@@ -202,6 +223,49 @@ func decodeListByOrg(_ context.Context, r *http.Request) (interface{}, error) {
...
 
+func decodeListByMetadata(_ context.Context, r *http.Request) (interface{}, error) {
+	pm, err := apiutil.BuildPageMetadataFromBody(r)
+	if err != nil {
+		return nil, err
+	}
+
+	req := listReq{
+		token:        apiutil.ExtractBearerToken(r),
+		pageMetadata: pm}
```
manuio: Same here

things/api/http/things/transport.go
```diff
@@ -279,6 +291,64 @@ func decodeListByOrg(_ context.Context, r *http.Request) (interface{}, error) {
...
 
+func decodeListByMetadata(_ context.Context, r *http.Request) (interface{}, error) {
+	pm, err := apiutil.BuildPageMetadataFromBody(r)
+	if err != nil {
+		return nil, err
+	}
+
+	req := listReq{
+		token:        apiutil.ExtractBearerToken(r),
+		pageMetadata: pm}
```
manuio: and here

manuio (changes requested):

pkg/apiutil/transport.go
Nesa002: If I use the PageMetadata struct directly, then I won't be able to differentiate between explicit zero values for limit or just ommited limit field.
They will both be handled as nil and we will lose the default behaviour when reading from query parameters that handles limit 0 as error and nil as default limit: https://github.com/MainfluxLabs/mainflux/blob/master/pkg/apiutil/transport.go#L170
Do you still want me to go through with the change?

manuio: Limit can't be 0. It would mean to return nothing.

auth/api/http/orgs/endpoint_test.go
```diff
@@ -599,6 +612,184 @@ func TestListOrgs(t *testing.T) {
...
 
+func TestSearchOrgs(t *testing.T) {
+	svc := newService()
+	_, token, err := svc.Issue(context.Background(), "", auth.Key{Type: auth.LoginKey, IssuedAt: time.Now(), IssuerID: id, Subject: email})
+	assert.Nil(t, err, fmt.Sprintf("Issuing login key expected to succeed: %s", err))
+
+	ts := newServer(svc)
+	defer ts.Close()
+
+	str := searchOrgReq
```
manuio: Rename to `sor`

auth/api/http/orgs/endpoint_test.go
```diff
@@ -599,6 +612,184 @@ func TestListOrgs(t *testing.T) {
...
+	validData := toJSON(str)
+
+	str.Dir = "desc"
+	str.Order = "name"
+	descData := toJSON(str)
+
+	str.Dir = "asc"
+	str.Order = "name"
```
manuio: This was already set in previous lines. Should we maybe use strings instead of PageMetadat? It would be more readable, avoid `toJSON` and it's consistent with `invalidData`.

auth/api/http/orgs/transport.go
```diff
@@ -57,6 +57,13 @@ func MakeHandler(svc auth.Service, mux *bone.Mux, tracer opentracing.Tracer, log
 		opts...,
 	))
 
+	mux.Post("/orgs/search", kithttp.NewServer(
```
manuio, line 60: Add the new endpoints in corresponding openapi files

manuio (changes requested):

auth/api/http/orgs/transport.go
```diff
@@ -88,6 +95,20 @@ func decodeListOrgs(_ context.Context, r *http.Request) (interface{}, error) {
...
+	}
+
+	req := listOrgsReq{
+		token:        apiutil.ExtractBearerToken(r),
+		pageMetadata: pm,
+	}
+
+	return req, nil
+}
+
```
majabirmancevic, line 111: can we use `decodeListOrgs` here, since the same request (listOrgsReq) is used?

consumers/notifiers/api/http/endpoint_test.go
```diff
@@ -744,3 +939,12 @@ func runRemoveNotifiersTest(t *testing.T, validContacts []string) {
 		assert.Equal(t, tc.status, res.StatusCode, fmt.Sprintf("%s: expected status code %d got %d", tc.desc, tc.status, res.StatusCode))
 	}
 }
+
+type SearchNotifiersRequest struct {
```
majabirmancevic: searchNotifiersRequest

things/api/http/groups/endpoint_test.go
```diff
@@ -26,24 +26,43 @@ import (
...
-	orgsList  = []auth.Org{{ID: orgID, OwnerID: user.ID}}
+	user            = users.User{ID: "574106f7-030e-4881-8ab0-151195c29f94", Email: userEmail, Password: password, Role: auth.Owner}
+	otherUser       = users.User{ID: "ecf9e48b-ba3b-41c4-82a9-72e063b17868", Email: otherUserEmail, Password: password, Role: auth.Editor}
+	admin           = users.User{ID: "2e248e36-2d26-46ea-97b0-1e38d674cbe4", Email: adminEmail, Password: password, Role: auth.RootSub}
+	group           = things.Group{Name: "test-group", Description: "test-group-desc", OrgID: orgID}
+	usersList       = []users.User{admin, user, otherUser}
+	orgsList        = []auth.Org{{ID: orgID, OwnerID: user.ID}}
+	metadata        = map[string]interface{}{"test": "data"}
+	invalidName     = strings.Repeat("m", maxNameSize+1)
+	invalidNameData = fmt.Sprintf(`{"limit":5,"offset":0,"name":"%s"}`, invalidName)
```
manuio: You can unify these lines

manuio (changes requested):

api/openapi/auth.yml
```diff
@@ -97,6 +97,26 @@ paths:
           description: Missing or invalid access token provided.
         '500':
           $ref: "#/components/responses/ServiceError"
+  /orgs/search:
+    post:
+      summary: Search and retrieve organizations.
+      description: |
+        Retrieves a list of organizations with optional name and metadata filtering.
```
manuio: There is also Order, Dir and others

anna02272 (approved):

manuio (approved):
LGTM

PR #733, opened 2025-07-10
MF-730 - Remove support for -1 limit to fetch all entities

### LINKED ISSUE #730: Remove -1 Limit filter
Currently we can pass -1 as Limit to fetch all entities without pagination. `ReadLimitQuery` (https://github.com/MainfluxLabs/mainflux/blob/master/pkg/apiutil/transport.go#L154) will set the Limit to 0 and the Limit query will be an empty string: https://github.com/MainfluxLabs/mainflux/blob/master/pkg/dbutil/query.go#L83

Fix `ReadLimitQuery` so only values bigger than 0 are accepted.

Nesa002:
Removed the option to fetch all entities by setting the limit to -1. All queries must use a defined limit. Resolves #730

### DIFF
--- a/consumers/notifiers/api/http/endpoint_test.go
+++ b/consumers/notifiers/api/http/endpoint_test.go
@@ -294,13 +294,6 @@ func runListNotifiersByGroupTest(t *testing.T, validContacts []string) {
 			res:    data[0:5],
 		},
 		{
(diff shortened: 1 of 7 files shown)


manuio (approved):
LGTM

PR #736, opened 2025-07-10
MF-732 - Return ErrGroupMembershipExists when membership is already created

### LINKED ISSUE #732: Return ErrGroupMembershipExists when membership is already created
Return `ErrGroupMembershipExists` if the membership already exists, as was done for org memberships:
https://github.com/MainfluxLabs/mainflux/blob/3d39cee5a5f00cf6555d2b3a6a022c72dcd464c6/auth/postgres/memberships.go#L119

majabirmancevic:
Resolves #732


manuio (approved):
LGTM

PR #737, opened 2025-07-11
NOISSUE - Remove references to deprecated Bootstrap service from Go SDK

bool3max:
Removes the `BootstrapURL` struct field in the SDK implementation, its `Config`, all all places that initialized those fields.


manuio (approved):
LGTM

PR #742, opened 2025-07-14
MF-731 - Use consistent database method names

### LINKED ISSUE #731: Use consistent database method names
Somewhere we use names like` RetrieveBy<entity><ID>`, for example [here](https://github.com/MainfluxLabs/mainflux/blob/3d39cee5a5f00cf6555d2b3a6a022c72dcd464c6/webhooks/webhooks.go#L34), but somewhere it's without 'ID'.
Since retrieving (and other operations) are performed based on IDs, it is enough to use only the entity within the name, without emphasizing its ID,
so instead of RetrieveByGroupID, RetrieveByGroupIDs, we would have RetrieveByGroup, RetrieveByGroups, etc..

majabirmancevic:
Resolves #731

### DIFF
--- a/auth/memberships.go
+++ b/auth/memberships.go
@@ -53,8 +53,8 @@ type OrgMembershipsRepository interface {
 	// RetrieveRole retrieves role of membership specified by memberID and orgID.
 	RetrieveRole(ctx context.Context, memberID, orgID string) (string, error)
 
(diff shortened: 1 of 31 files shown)


Nesa002 (approved):

manuio (approved):
LGTM

PR #743, opened 2025-07-14
MF-740- Remove query parameter from delete messages endpoint and enhanced table handling

### LINKED ISSUE #740: Delete messages deleting messages from both tables and having redundant query parameter
Currently, delete messages endpoint deletes messages from both tables which needs to be specified. Also, `publisherID` shouldn't be in query parameter but extracted.

duolok:
Resolves #740


readers/api/endpoint.go
```diff
@@ -46,6 +46,7 @@ func listAllMessagesEndpoint(svc readers.MessageRepository) endpoint.Endpoint {
 				return nil, err
 			}
 			req.pageMeta.Publisher = pc.PublisherID
+			pc.ProfileConfig.GetContentType()
```
manuio: Remove this line

manuio (changes requested):

readers/api/transport.go
```diff
@@ -198,9 +195,8 @@ func decodeDeleteMessages(_ context.Context, r *http.Request) (interface{}, erro
 		token: apiutil.ExtractBearerToken(r),
 		key:   apiutil.ExtractThingKey(r),
 		pageMeta: readers.PageMetadata{
-			Publisher: publisherID,
-			From:      from,
-			To:        to,
+			From:        from,
```
manuio: Fix the indentation

pkg/dbutil/query.go
```diff
@@ -77,6 +77,17 @@ func GetOffsetLimitQuery(limit uint64) string {
 	return ""
 }
 
+func GetDeleteTable(format string) string {
```
majabirmancevic: GetTableName?

manuio (changes requested):

pkg/dbutil/query.go
```diff
@@ -77,6 +77,17 @@ func GetOffsetLimitQuery(limit uint64) string {
...
+	switch format {
+	case "json":
```
majabirmancevic: use consts `messaging.JSONFormat`, `messaging.SenMLFormat`

readers/api/endpoint.go
```diff
@@ -88,10 +88,17 @@ func deleteMessagesEndpoint(svc readers.MessageRepository) endpoint.Endpoint {
 			if err != nil {
 				return nil, errors.Wrap(errors.ErrAuthentication, err)
 			}
+			req.pageMeta.Publisher = pc.PublisherID
 
-			if pc.PublisherID != req.pageMeta.Publisher {
-				return nil, errors.ErrAuthentication
+			switch pc.ProfileConfig.GetContentType() {
```
majabirmancevic: check this func: https://github.com/MainfluxLabs/mainflux/blob/7c664b6668b77e03b555fedcd7d61ae787f42d94/pkg/messaging/nats/publisher.go#L74
and use it here, to avoid duplication

readers/api/endpoint.go
```diff
@@ -89,15 +90,7 @@ func deleteMessagesEndpoint(svc readers.MessageRepository) endpoint.Endpoint {
...
-
-			switch pc.ProfileConfig.GetContentType() {
-			case "application/senml+json":
-				req.pageMeta.Format = "senml"
-			case "application/json":
-				req.pageMeta.Format = "json"
-			default:
-				req.pageMeta.Format = "senml"
-			}
+			req.pageMeta.Format = nats.GetFormat(pc.ProfileConfig.GetContentType())
```
majabirmancevic: probably it makes more sense to move GetFormat to [messaging/pubsub](https://github.com/MainfluxLabs/mainflux/blob/master/pkg/messaging/pubsub.go), because we don't use it only in nats anymore @manuio

manuio: Right but I'd maybe create a new method that accept a CT and convert to the table name. We are converting twice. In any case, I wouldn't use messaging package here.

majabirmancevic: set table name to `req.pageMeta.Format`?

manuio: Let's use it as argument and remove Format from pageMeta even for reading.

pkg/messaging/nats/publisher.go
```diff
@@ -60,7 +60,7 @@ func GetSubjects(pc *protomfx.Config, subtopic string) []string {
 	subjects := []string{subjectWebhook}
 
 	if pc.GetWrite() {
-		format := getFormat(pc.ContentType)
+		format := GetFormat(pc.ContentType)
```
manuio: revert it

manuio (changes requested):

readers/postgres/messages_test.go
```diff
@@ -585,12 +585,14 @@ func TestDeleteMessagesSenML(t *testing.T) {
 			Publisher: pubID,
 			From:      0,
 			To:        now,
-		})
+		}, "messages")
```
manuio: Use a const here

readers/postgres/messages_test.go
```diff
@@ -784,13 +786,13 @@ func TestDeleteMessagesJSON(t *testing.T) {
 			Publisher: id1,
 			From:      0,
 			To:        float64(created + int64(msgsNum)),
-		})
+		}, "json")
```
manuio: and here

readers/postgres/messages_test.go
```diff
@@ -585,12 +585,14 @@ func TestDeleteMessagesSenML(t *testing.T) {
...
 			To:        now,
-		})
+		}, "messages")
+
 		_ = reader.DeleteMessages(context.Background(), readers.PageMetadata{
 			Publisher: pubID2,
 			From:      0,
 			To:        now,
-		})
+		}, "messages")
```
manuio: and here

readers/postgres/messages_test.go
```diff
@@ -809,7 +811,7 @@ func TestDeleteMessagesJSON(t *testing.T) {
 		require.Nil(t, err)
 		beforeCount := beforePage.Total
 
-		err = reader.DeleteMessages(context.Background(), tc.pageMeta)
+		err = reader.DeleteMessages(context.Background(), tc.pageMeta, "json")
```
manuio: and here

manuio (changes requested):

readers/postgres/messages_test.go
```diff
@@ -32,6 +32,8 @@ const (
 	msgName     = "temperature"
 	jsonFormat  = "json"
```
manuio, line 33: This one is duplicated but out of scope in this PR. Just keep it in mind.

manuio (approved):
LGTM

majabirmancevic (approved):

Issue #746, opened 2025-07-15
Require e-mail verification during the user self-register process


bool3max:
The current implementation of the self-register process (i.e. a person registering themselves a new account on the platform without the need for any existing users/admins to take action) doesn't carry out any verification against the e-mail address provided by the user.

We should require the user to *verify* the e-mail address they claim to have access to. This has numerous benefits, and could be implemented using the well known paradigm of sending a URL containing a unique ID associated with a particular account registration process to the e-mail in question. Once the user navigates to the URL, the e-mail can be considered verified and the registration process complete.

bool3max:
Resolved by #759

PR #752, opened 2025-07-16
MF-706 - Implement Org and Platform Invites

### LINKED ISSUE #706: Implement Organization membership e-mail invites
The current Organization member-invite flow is very bare-bones:

1. A Mainflux user with "admin" privileges within a certain Organization decides to add another User as a member of the Org
2. They enter the e-mail of the User they wish to invite, also choosing an appropriate member role
3. If the e-mail belongs to an existing Mainflux User, that user is "**silently**" added as a member of the Org, with an appropriate role
4. If the e-mail does not belong to any existing Mainflux User, the Org admin is presented with an error

### Invites

The idea is to present the invitee (user being invited) with an e-mail notification containing information about the invitation. 

Here's what I currently have in mind regarding the above-described invite flow, enhanced with support for e-mail notifications:

1. A Mainflux user with "admin" privileges within a certain Organization decides to add another User as a member of the Org
2. They enter the-mail of the User they wish to invite, choosing an appropriate member role as well
3. If the e-mail belongs to an existing Mainflux user, they are sent an e-mail containing detailed information about the invite (inviter e-mail, Organization name, assigned role, time of invite, etc...), alongside a URL to the Mainflux dashboard where they'd be able to view the invite once more

    - **QUESTION**: Should we support the ability for the user to "decline" the invitation? It seems logical, otherwise users could get spammed and unconditionally added to Orgs without their consent.
4. **QUESTION**: If the e-mail does not belong to any existing User, should they get a chance to register a new account first? -- In other words, the user would still get an appropriate invitation e-mail, but the URL within would first lead to a registration page where the new User would complete the registration process, after which they'd be redirected to the invitation where they'd potentially be able to accept/decline it.

---

This Issue will serve as a discussion space about the implementation.

bool3max:
This PR implements the concept of Platform and Org invites as discussed in Issue #706.

# Concept

#706 includes an early idea and some discussion of how Invites should work. Below is an explanation of the concepts and how they are implemented in this PR *currently*, as well as a list of questions and uncertainties that we have yet to decide on:

## Org Invites

* Org invites live in the `auth` service.
* An Org Invite is a request for a specific User to join a specific Org, created and sent on behalf an existing Org member with a role of `admin` (or higher).
    * Invites are always created and sent based on the invitee user's email.
    * The e-mail must belong to an existing, registered Mainflux user
* The same Inviter cannot send multiple Invites to the same Invitee to join the same Organization
* A *pending* Invite can be revoked at any time by the Inviter (i.e. User who sent it)
* Once an Invite is sent to a User, they also receive an e-mail notification:
    * The content of this e-mail notification contains basic info about the Invite: Organization name, e-mail of the Inviter (i.e. "who invited me"), time of invite, time of expiration, and crucially, a link to the Mainflux UI dashboard, where the user can inspect the Invite once more, and potentially **respond** to it
* Users can view a list of all pending Invites sent to them at any time by using the appropriate API endpoint. The idea is for this to become a page in the UI dashboard.
* Users can also view a list of all Invites *sent out* by them. Similarly this can probably eventually become a tab in the UI dashboard, from which the users would quickly be able to *revoke* invites.
* When **responding** to an Invite, the Invitee the can choose to either **accept** or **decline** the invite.
    * By accepting the Invite, the Invitee is immediately added as a member of the corresponding Org, and the corresponding Invite is marked as `accepted`
    * By declining the Invite, the invite is marked as *declined* and no further action is taken
* **Org Invites expire**. 
    * This is a security measure.
    * The default expiration time is 7 days, but it is configurable via an environment variable.
    * Attempting to *interact* with an expired Invite in any way using its ID, for example trying to view or respond to one such Invite always results in an appropriate error.

## Platform Invites

* Platform invites live in the `users` service.
* A Platform Invite is an invite made on behalf of the platform Root Admin for an unregistered user to create an account on the platform. It is created and sent based on the user's email address.
    * The e-mail address may not belong to an existing Mainflux user account. 
    * All operations concerning platform invites (creating, listing, revoking, etc...) are only possible when authenticated as the Root Admin user.
* The recipient, i.e. invitee of a platform invite may *accept* it by making a request to a special platform-invite-only register API endpoint (see table below), submitting the invite ID, and the usual user-registration information (e-mail address, password, etc...).
    * The submitted invite ID must belong to an existing, valid, non-expired platform invite, and the submitted e-mail address of the new account must match the e-mail address associated with that platform invite. These checks ensure that the user can only register an account on the platform if they've received an invite to do so. 
    * This endpoint is separate from the existing self-register (`POST /register`) and unconditional admin-only registration (`POST /users`) endpoints.
* Platform invites also expire.
    * The default expiration time is 7 days, but it is configurable via an environment variable.

# HTTP API endpoints

Below is a list of all new HTTP API endpoints introduced in the `auth` and `users` services:

## `auth` service - Org invites

| Method | Endpoint | Description |
|---|---|---|
| POST | `/orgs/:id/invites` | Create a new Org Invite. |
| GET | `/orgs/:id/invites` | Retrieve a list of all Org Invites for a specific Org. |
| GET | `/invites/:id` | Retrieve information about a specific Org Invite. |
| DELETE | `/invites/:id` | Revoke ("remove") a specific Org Invite. |
| POST | `/invites/:id/:action` | Respond to an Org Invite. `action` must be one of `accept` or `decline`. |
| GET | `/users/:userID/invites/received` | Retrieve a list of Invites *received* by a specific User (invitee). |
| GET | `/users/:userID/invites/sent` | Retrieve a list of Invites *sent* by a specific User (inviter). |

## `users` service - Platform invites

| Method | Endpoint | Description |
|---|---|---|
| POST | `/invites` | Create a new Platform Invite. Usable only by Root Admin. |
| GET | `/invites` | Retrieve a list of all Platform Invites. Usable only by Root Admin. |
| GET | `/invites/:id` | Retrieve information about a specific Platform Invite. Usable only by Root Admin. |
| DELETE | `/invites/:id` | Revoke ("remove") a specific Platform Invite. Usable only by Root Admin. |
| POST | `/register/invite/:id` | Register a new user account using a pending Platform Invite. |

# TODO

* [x] Implement ability to invite new users who do not yet have a Mainflux user account
    * This depends on ironing out and improving existing self-register functionality. See e.g. #746 
* [x] Implement SDK methods for new functionality
* [ ] Implement CLI commands
* [x] Implement tests for all new functionality:
    * [x] `auth` service method tests
    * [x] `auth` HTTP API endpoint tests
    * [x] `auth` Postgres `invites` tests
* [ ] Documentation
    * [ ] `MainfluxLabs/docs` documentation
    * [x] OpenAPI


bool3max:
Regarding inviting unregistered users: we could it implement it one of a couple ways. Here's what I have in mind:

1. When inviting a User whose email does not belong to a Mainflux user, simply send them an e-mail saying that they've been invited to a Mainflux Organization, but that they first need to create an account, and provide them with a registration link.
    * This is the simplest option as the user isn't actually "invited" the Org at all. Once they create an account, they'd need to be invited *again*.
2. When inviting a User whose email does not belong to a Mainflux account, temporarily keep track of the yet-to-be-created invitation in the database alongside a unique token, and send the user an e-mail containing an account registration link with the unique token embedded within. During the registration process also forward out the token from the e-mail, and if it matches an existing one in the database, also automatically send out the appropriate Invite, *or* simply assign them to the Org immediately (?).
    * This would be pretty seamless from the end-user point of view, but would be a bit more complex to implement as we'd have to slightly modify the way registration works on both the backend and the frontend.

auth/api/http/invites/requests.go
```diff
@@ -0,0 +1,118 @@
...
+func (req listOrgInvitesByUserReq) validate() error {
+	if req.token == "" {
+		return apiutil.ErrBearerToken
+	}
+
+	if req.userID == "" {
+		return apiutil.ErrMissingUserID
+	}
+
+	if err := apiutil.ValidatePageMetadata(req.pm, maxLimitSize, 254); err != nil {
```
anna02272: use const `maxNameSize  = 254`

auth/api/http/invites/requests.go
```diff
@@ -0,0 +1,118 @@
...
+func (req listOrgInvitesByOrgReq) validate() error {
+	if req.token == "" {
+		return apiutil.ErrBearerToken
+	}
+
+	if req.orgID == "" {
+		return apiutil.ErrMissingOrgID
+	}
+
+	if err := apiutil.ValidatePageMetadata(req.pm, maxLimitSize, 254); err != nil {
```
anna02272: use const `maxNameSize  = 254`

users/api/http/requests.go
```diff
@@ -216,6 +229,62 @@ func (req changeUserStatusReq) validate() error {
...
+	token string
+	pm    apiutil.PageMetadata
+}
+
+func (req listPlatformInvitesRequest) validate() error {
+	if req.token == "" {
+		return apiutil.ErrBearerToken
+	}
+
+	if err := apiutil.ValidatePageMetadata(req.pm, maxLimitSize, 254); err != nil {
```
anna02272: use const `maxNameSize  = 254`

api/openapi/auth.yml
```diff
@@ -698,6 +863,78 @@ components:
...
+          description: The organization's unique identifier
+        invitee_role:
+          type: string
+          description: The invitee's proposed role within the organization.
+        created_at:
+          type: string
+          format: date-time
+          example: "2024-05-16 16:05:01"
+          description: Time at which the Invite was created.
+        updated_at:
```
anna02272: expires_at?

bool3max: Yup!

api/openapi/users.yml
```diff
@@ -173,6 +194,80 @@ paths:
           description: Missing or invalid access token provided.
         '500':
           $ref: "#/components/responses/ServiceError"
+  /invites-platform:
+    post:
+      summary: Create new Platform Invite.
+      description: |
+        Create a new Platform Invite towards an unregistered user to createa a Mainflux user account.
```
anna02272: create a Mainflux user account.

api/openapi/users.yml
```diff
@@ -499,6 +618,47 @@ components:
...
+          description: E-mail of invited user.
+        created_at:
+          type: string
+          format: date-time
+          example: "2024-05-16 16:05:01"
+          description: Time at which the Invite was created.
+        expires_at:
+          type: string
+          format: date-time
+          example: "2024-05-2316:05:01"
```
anna02272: "2024-05-23 16:05:01"

auth/api/http/invites/endpoint_test.go
```diff
@@ -0,0 +1,593 @@
...
+		{
+			desc:   "list invites with empty auth token",
+			url:    fmt.Sprintf("%s/users/%s/invites/received", ts.URL, viewerID),
+			token:  "",
+			status: http.StatusUnauthorized,
+			res:    nil,
+		},
+		{
+			desc:   "list invites with negative offset",
+			url:    fmt.Sprintf("%s/users/%s/invites/received?offset=%d/received", ts.URL, viewerID, -1),
```
anna02272: "%s/users/%s/invites/received?offset=%d" ?

auth/api/http/invites/transport.go
```diff
@@ -0,0 +1,213 @@
...
+	kitot "github.com/go-kit/kit/tracing/opentracing"
+	kithttp "github.com/go-kit/kit/transport/http"
+	"github.com/go-zoo/bone"
+	"github.com/opentracing/opentracing-go"
+)
+
+const (
+	userIDKey            = "userID"
+	inviteIDKey          = "inviteID"
+	inviteReponseVerbKey = "responseVerb"
```
anna02272: Typo: inviteRe**s**ponseVerbKey

users/emailer/metrics.go
```diff
@@ -41,3 +41,12 @@ func (ms *metricsMiddleware) SendEmailVerification(To []string, redirectPath, to
 
 	return ms.emailer.SendEmailVerification(To, redirectPath, token)
 }
+
+func (ms *metricsMiddleware) SendPlatformInvite(To []string, invite users.PlatformInvite, redirectPath string) error {
+	defer func(begin time.Time) {
+		ms.counter.With("method", "send_email_verification").Add(1)
```
anna02272: send_platform_invite

users/emailer/metrics.go
```diff
@@ -41,3 +41,12 @@ func (ms *metricsMiddleware) SendEmailVerification(To []string, redirectPath, to
...
+		ms.latency.With("method", "send_email_verification").Observe(time.Since(begin).Seconds())
```
anna02272: send_platform_invite

auth/api/http/invites/requests.go
```diff
@@ -0,0 +1,121 @@
...
+
+const (
+	maxLimitSize = 200
+	maxNameSize  = 254
+)
+
+type createOrgInviteReq struct {
+	token        string
+	orgID        string
+	OrgMember    auth.OrgMembership `json:"org_member,omitempty"`
```
manuio: Shouldn't this be `OrgMembership`?

auth/api/http/invites/transport.go
```diff
@@ -0,0 +1,234 @@
...
+
+	inviteResponseVerb := bone.GetValue(r, inviteResponseVerbKey)
+	switch inviteResponseVerb {
+	case "accept":
+		req.inviteAccepted = true
+	case "decline":
+		req.inviteAccepted = false
+	default:
+		return orgInviteResponseReq{}, apiutil.ErrInvalidInviteResponse
+	}
```
manuio: If we are using the same in users service this could be an apiutil method

auth/api/http/invites/requests.go
```diff
@@ -0,0 +1,121 @@
...
+		return apiutil.ErrMissingInviteID
+	}
+
+	return nil
+}
+
+type orgInviteResponseReq struct {
+	token          string
+	inviteID       string
+	inviteAccepted bool
```
manuio: Should we simply use `accepted`?

auth/api/http/invites/responses.go
```diff
@@ -0,0 +1,114 @@
...
+func (res respondOrgInviteRes) Headers() map[string]string {
+	return map[string]string{}
+}
+
+func (res respondOrgInviteRes) Empty() bool {
+	return true
+}
+
+type orgInviteRes struct {
+	ID          string    `json:"id"`
```
manuio, line 71: I guess that this is `InviteID`. Should we use `ID` for other request and response structures?

auth/api/http/invites/transport.go
```diff
@@ -0,0 +1,234 @@
...
+		token: apiutil.ExtractBearerToken(r),
+		orgID: bone.GetValue(r, apiutil.IDKey),
+	}
+
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		return nil, errors.Wrap(apiutil.ErrMalformedEntity, err)
+	}
+
+	if req.OrgMember.Role == "" {
+		req.OrgMember.Role = auth.Viewer
```
manuio: Return an error if the API isn't well implemented

auth/mocks/users.go
```diff
@@ -65,6 +68,8 @@ func (svc *usersServiceClientMock) GetUsersByEmails(_ context.Context, in *proto
 	for _, email := range in.Emails {
 		if user, ok := svc.usersByEmails[email]; ok {
 			users = append(users, &protomfx.User{Id: user.ID, Email: user.Email})
+		} else {
+			return nil, status.Error(codes.NotFound, errors.ErrNotFound.Error())
```
manuio: Avoid if else. Check if `!ok` and return an error instead.

auth/postgres/init.go
```diff
@@ -96,6 +96,32 @@ func migrateDB(db *sqlx.DB) error {
...
+						created_at   TIMESTAMPTZ,
+						expires_at   TIMESTAMPTZ,
+						state        VARCHAR DEFAULT 'pending' NOT NULL,      
+						FOREIGN KEY  (org_id) REFERENCES orgs (id) ON DELETE CASCADE,
+						PRIMARY KEY  (id)
+					)
+					`,
+					`
+					CREATE UNIQUE INDEX ux_invites_org_invitee_id_org_id on invites_org (invitee_id, org_id) WHERE state='pending'
+					`,
```
manuio: Fix the indentation and use the same format than other migrations

auth/api/http/invites/requests.go
```diff
@@ -0,0 +1,121 @@
...
+	if req.OrgMember.Role != auth.Admin && req.OrgMember.Role != auth.Viewer && req.OrgMember.Role != auth.Editor {
+		return apiutil.ErrInvalidRole
+	}
+
+	return nil
+}
+
+type inviteReq struct {
+	token    string
+	inviteID string
```
manuio: Why `inviteID` and not `ID`?

auth/postgres/init.go
```diff
@@ -96,6 +96,32 @@ func migrateDB(db *sqlx.DB) error {
...
+						invitee_role VARCHAR(12) NOT NULL,
+						created_at   TIMESTAMPTZ,
+						expires_at   TIMESTAMPTZ,
+						state        VARCHAR DEFAULT 'pending' NOT NULL,      
+						FOREIGN KEY  (org_id) REFERENCES orgs (id) ON DELETE CASCADE,
+						PRIMARY KEY  (id)
+					)
+					`,
+					`
+					CREATE UNIQUE INDEX ux_invites_org_invitee_id_org_id on invites_org (invitee_id, org_id) WHERE state='pending'
```
manuio: Can we define it next to the table? 
```
            PRIMARY KEY  (id),
            CONSTRAINT   ux_invites_org_invitee_id_org_id 
                         UNIQUE (invitee_id, org_id) 
                         WHERE state = 'pending'
```

Also, can we simplify the name?

auth/postgres/invites.go
```diff
@@ -0,0 +1,428 @@
...
+			tx.Rollback()
+
+			pgErr, ok := err.(*pgconn.PgError)
+			if ok {
+				switch pgErr.Code {
+				case pgerrcode.InvalidTextRepresentation:
+					return errors.Wrap(errors.ErrMalformedEntity, err)
+				case pgerrcode.UniqueViolation:
+					var e = errors.ErrConflict
+					if pgErr.ConstraintName == "ux_invites_org_invitee_id_org_id" {
```
manuio: What other type of conflicts can we have?

auth/postgres/invites.go
```diff
@@ -0,0 +1,428 @@
...
+	if err := ir.db.QueryRowxContext(ctx, q, inviteID).StructScan(&dbI); err != nil {
+		if err == sql.ErrNoRows {
+			return auth.OrgInvite{}, errors.Wrap(errors.ErrNotFound, err)
+		}
+
+		pgErr, ok := err.(*pgconn.PgError)
+		if ok {
+			switch pgErr.Code {
+			case pgerrcode.InvalidTextRepresentation:
+				return auth.OrgInvite{}, errors.Wrap(errors.ErrMalformedEntity, err)
```
manuio: We should use another error. We should also remove `errors.ErrMalformedEntity` in favour to apiutil.ErrMalformedEntity. This error is used to indicate a malformed request not for the DB layer.

manuio (changes requested):

auth/api/http/invites/requests.go
bool3max: Yup, I'll rename it.

auth/api/http/invites/transport.go
bool3max: We're not - platform invites (in the users service) cannot be "declined" - only accepted.

auth/api/http/invites/requests.go
bool3max: I agree, it's better. I'll rename it.

auth/api/http/invites/responses.go
bool3max: Yes when it's clear that the struct represents a single type of entity the field can be just `ID`. I'll clean up the others to follow.

auth/api/http/invites/transport.go
bool3max: It's done the same way in the endpoint for creating Org Memberships:

https://github.com/MainfluxLabs/mainflux/blob/066a8cbebed379be63f7177e18ff17fb88b591b3/auth/api/http/memberships/transport.go#L122

.. So I repeat the logic here. Should we change it in both?

auth/api/http/invites/requests.go
bool3max: Yes just `ID` makes more sense, I'll change it

auth/postgres/invites.go
bool3max: None :) - it's a remnant of when there were two possible constraints -- I'll remove the check

auth/postgres/init.go
bool3max: Right, we can embed it in the table definition itself. I'm not sure about the name, I tried to look up a convention for naming the indexes but couldn't find anything concrete, so I just landed on a name that combines the names of the table and the columns it enforces.

Maybe `single_invite_per_invitee` or something?

auth/api/http/invites/transport.go
manuio: ok

auth/api/http/invites/transport.go
manuio: Yes please, better to force users to implement properly the API.

auth/postgres/init.go
manuio: Let's use `unique_org_invitee_pending`

bool3max: @manuio Actually it seems that creating a partial unique index inside a CREATE TABLE statement is not supported, so it has to be a separate string in the migration slice.

pkg/apiutil/errors.go
```diff
@@ -54,6 +54,12 @@ var (
 	// ErrMissingKeyID indicates missing ID of key.
 	ErrMissingKeyID = errors.New("missing key ID")
 
+	// ErrMissingInviteID incidates missing ID of Invite.
+	ErrMissingInviteID = errors.New("missing invite ID")
+
+	// ErrInvalidInviteResponse indicates an invalid Invite response verb string.
+	ErrInvalidInviteResponse = errors.New("invalid invite response verb")
```
bool3max: This error can be moved to the `auth` service as it's only used there.

pkg/apiutil/errors.go
```diff
@@ -164,4 +170,16 @@ var (
 
 	// ErrInvalidOperator indicates an invalid logical operator
 	ErrInvalidOperator = errors.New("missing or invalid logical operator")
+
+	// ErrCreateInvite indicates failure to create a new Invite
+	ErrCreateInvite = errors.New("error creating invite")
```
bool3max: This error does not need to exist at all, just return `dbutil.ErrCreateEntity` where appropriate.

users/invites.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	go func() {
+		svc.SendPlatformInviteEmail(ctx, invite, redirectPath)
+	}()
+
+	return invite, nil
+}
+
+func (svc usersService) RevokePlatformInvite(ctx context.Context, token string, inviteID string) error {
```
manuio: `token, inviteID string`

auth/api/logging.go
```diff
@@ -313,3 +313,101 @@ func (lm *loggingMiddleware) RetrieveRole(ctx context.Context, id string) (role
...
+			return
+		}
+
+		lm.logger.Info(fmt.Sprintf("%s without errors.", message))
+	}(time.Now())
+
+	return lm.svc.InviteOrgMember(ctx, token, orgID, invRedirectPath, om)
+}
+
+func (lm *loggingMiddleware) RevokeOrgInvite(ctx context.Context, token string, inviteID string) (err error) {
```
manuio: `token, inviteID string`

auth/api/logging.go
```diff
@@ -313,3 +313,101 @@ func (lm *loggingMiddleware) RetrieveRole(ctx context.Context, id string) (role
...
+			return
+		}
+
+		lm.logger.Info(fmt.Sprintf("%s without errors.", message))
+	}(time.Now())
+
+	return lm.svc.RevokeOrgInvite(ctx, token, inviteID)
+}
+
+func (lm *loggingMiddleware) RespondOrgInvite(ctx context.Context, token string, inviteID string, accept bool) (err error) {
```
manuio: Same here and for all others

auth/api/metrics.go
```diff
@@ -224,3 +224,66 @@ func (ms *metricsMiddleware) RetrieveRole(ctx context.Context, id string) (strin
...
+func (ms *metricsMiddleware) RespondOrgInvite(ctx context.Context, token string, inviteID string, accept bool) error {
+	defer func(begin time.Time) {
+		ms.counter.With("method", "respond_org_invite").Add(1)
+		ms.latency.With("method", "respond_org_invite").Observe(time.Since(begin).Seconds())
+	}(time.Now())
+
+	return ms.svc.RespondOrgInvite(ctx, token, inviteID, accept)
+}
+
+func (ms *metricsMiddleware) ListOrgInvitesByUser(ctx context.Context, token string, userType string, userID string, pm auth.PageMetadataInvites) (auth.OrgInvitesPage, error) {
```
manuio: `oken, userType, userID string`

manuio (changes requested):

api/openapi/auth.yml
```diff
@@ -346,6 +346,171 @@ paths:
...
+          description: Unauthorized.
+        '404':
+          description: Failed due to non existing Invite.
+        '500':
+          $ref: "#/components/responses/ServiceError"
+  /users/{userId}/invites/received:
+    get:
+      summary: Retrieve pending Invites sent to specific user.
+      description: |
+        Retreieve a list of pending Invites send to a specific user, i.e. invitee, denoted by their
```
anna02272: Retrieve -typo

api/openapi/auth.yml
```diff
@@ -346,6 +346,171 @@ paths:
...
+          description: Missing or invalid access token provided.
+        '403':
+          description: Failed to perform authorization over the entity.
+        '500':
+          $ref: "#/components/responses/ServiceError"
+  /users/{userId}/invites/sent:
+    get:
+      summary: Retrieve pending Invites sent by a specific user.
+      description: |
+        Retreieve a list of pending Invites send by a specific user, i.e. inviter, denoted by their
```
anna02272: Retrieve -typo

auth/api/http/invites/requests.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	if req.id == "" {
+		return apiutil.ErrMissingInviteID
+	}
+
+	return nil
+}
+
+type orgInviteResponseReq struct {
```
majabirmancevic: or `respondOrgInviteReq`

api/openapi/auth.yml
```diff
@@ -346,6 +346,171 @@ paths:
...
+        $ref: "#/components/requestBodies/InviteReq"
+      responses:
+        '201':
+          description: Invite created.
+        '400':
+          description: Failed due to malformed JSON.
+        '401':
+          description: Missing or invalid access token provided.
+        '403':
+          description: Unauthorized.
```
anna02272: We could use "Failed to perform authorization over the entity" consistently across all endpoints. Right now, some responses use this, while others just say "Unauthorized" in this file.

api/openapi/auth.yml
```diff
@@ -346,6 +346,171 @@ paths:
...
+        Retrieve a list of all existing Invites for any user to join the Organization identified by its ID.
+      tags:
+        - org invites
+      parameters:
+        - $ref: "#/components/parameters/OrgId"
+      responses:
+        '200':
+          $ref: "#/components/responses/InvitesPageRes"
+        '403':
+          description: Unauthorized.
```
anna02272: `Failed to perform authorization over the entity`

api/openapi/auth.yml
```diff
@@ -346,6 +346,171 @@ paths:
...
+        - org invites
+      parameters:
+          - $ref: "#/components/parameters/InviteId"
+      responses:
+        '200':
+          $ref: "#/components/responses/InviteRes"
+        '404':
+          description: Failed due to non existing Invite.
+        '403':
+          description: Unauthorized.
```
anna02272: `Failed to perform authorization over the entity`

api/openapi/auth.yml
```diff
@@ -346,6 +346,171 @@ paths:
...
+        Inviter, i.e. the user who sent it.
+      tags:
+        - org invites
+      parameters:
+        - $ref: "#/components/parameters/InviteId"
+      responses:
+        '204':
+          description: Invite revoked.
+        '403':
+          description: Unauthorized.
```
anna02272: `Failed to perform authorization over the entity`

api/openapi/auth.yml
```diff
@@ -346,6 +346,171 @@ paths:
...
+        i.e. the user being invited.
+      tags:
+        - org invites
+      parameters:
+        - $ref: "#/components/parameters/InviteId"
+      responses:
+        '201':
+          description: Invite accepted and invitee added as a member of the organization.
+        '403':
+          description: Unauthorized.
```
anna02272: `Failed to perform authorization over the entity`

api/openapi/auth.yml
```diff
@@ -346,6 +346,171 @@ paths:
...
+        - org invites
+      parameters:
+        - $ref: "#/components/parameters/InviteId"
+      responses:
+        '204':
+          description: Invite declined and removed from the database.
+        '401':
+          description: Missing or invalid access token provided.
+        '403':
+          description: Unauthorized.
```
anna02272: `Failed to perform authorization over the entity`

auth/api/http/invites/endpoint.go
```diff
@@ -0,0 +1,162 @@
...
+			pageRes: pageRes{
+				Limit:  page.Limit,
+				Offset: page.Offset,
+				Total:  page.Total,
+			},
+			Invites: []orgInviteRes{},
+		}
+
+		for _, inv := range page.Invites {
+			resInv := orgInviteRes{
```
majabirmancevic: can we use a helper like `BuildOrgInviteRes`? 
(and in other endpoints as well)

auth/roles.go
```diff
@@ -1,6 +1,8 @@
 package auth
 
-import "context"
+import (
+	"context"
```
majabirmancevic: revert inline import?

auth/invites.go
```diff
@@ -0,0 +1,334 @@
...
+	InviteStateExpired  = "expired"
+	InviteStateRevoked  = "revoked"
+	InviteStateAccepted = "accepted"
+	InviteStateDeclined = "declined"
+)
+
+type Invites interface {
+	// InviteOrgMember creates a pending Invite on behalf of the User authenticated by `token`,
+	// towards the user in `om`, to join the Org identified by `orgID` with an appropriate role.
+	InviteOrgMember(ctx context.Context, token, orgID, invRedirectPath string, om OrgMembership) (OrgInvite, error)
```
majabirmancevic: `CreateOrgInvite`?

majabirmancevic (changes requested):

auth/api/http/invites/requests.go
```diff
@@ -0,0 +1,125 @@
...
+
+const (
+	maxLimitSize = 200
+	maxNameSize  = 254
+)
+
+type createOrgInviteReq struct {
+	token         string
+	orgID         string
+	OrgMembership auth.OrgMembership `json:"org_membership,omitempty"`
```
majabirmancevic: I'd use `email` and `role` as fields instead of the `OrgMembership` object, unless we expect some more fields from `OrgMembership`?

auth/api/http/invites/transport.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		return nil, errors.Wrap(apiutil.ErrMalformedEntity, err)
+	}
+
+	return req, nil
+}
+
+func decodeOrgInviteResponseRequest(_ context.Context, r *http.Request) (any, error) {
```
majabirmancevic: decodeRespondOrgInviteRequest should be clearer, right?

auth/api/http/invites/transport.go
```diff
@@ -0,0 +1,209 @@
...
+	kitot "github.com/go-kit/kit/tracing/opentracing"
+	kithttp "github.com/go-kit/kit/transport/http"
+	"github.com/go-zoo/bone"
+	"github.com/opentracing/opentracing-go"
+)
+
+const (
+	userIDKey             = "userID"
+	inviteIDKey           = "inviteID"
+	inviteResponseVerbKey = "responseVerb"
```
majabirmancevic: can we find a better name for "verb"? maybe `action`/`type` or something similar?

auth/api/http/invites/transport.go
```diff
@@ -0,0 +1,209 @@
...
+	))
+
+	mux.Get("/orgs/:id/invites", kithttp.NewServer(
+		kitot.TraceServer(tracer, "list_org_invites_by_org")(listOrgInvitesByOrgEndpoint(svc)),
+		decodeListOrgInvitesByOrgRequest,
+		encodeResponse,
+		opts...,
+	))
+
+	mux.Get("/invites/:inviteID", kithttp.NewServer(
```
majabirmancevic: `/invites/:id`

auth/api/http/invites/transport.go
```diff
@@ -0,0 +1,209 @@
...
+	))
+
+	mux.Post("/invites/:inviteID/:responseVerb", kithttp.NewServer(
+		kitot.TraceServer(tracer, "respond_org_invite")(respondOrgInviteEndpoint(svc)),
+		decodeOrgInviteResponseRequest,
+		encodeResponse,
+		opts...,
+	))
+
+	mux.Get("/users/:userID/invites/received", kithttp.NewServer(
```
majabirmancevic: `/users/:id/..`
when we have only one `id` parameter in the path, then simply `:id` is used bcs it's clear which entity it is based on /invites/, /users/ ..
and when we have more than one ID in the path, then we define specific IDs: `/users/:userID/invites/inviteID`, .. bcs of `bone.GetValue`

auth/api/http/invites/transport.go
bool3max: Yeah `ResponseRequest` looks kinda confusing

auth/api/http/invites/transport.go
bool3max: `action` sounds good to me!

auth/api/http/invites/transport.go
```diff
@@ -17,9 +17,9 @@ import (
 )
 
 const (
-	userIDKey             = "userID"
-	inviteResponseVerbKey = "responseVerb"
-	stateKey              = "state"
+	userIDKey               = "userID"
+	inviteResponseActionKey = "action"
```
majabirmancevic: or even `responseActionKey`

users/invites.go
```diff
@@ -0,0 +1,200 @@
...
+	if err := svc.isAdmin(ctx, token); err != nil {
+		return PlatformInvite{}, err
+	}
+
+	_, err := svc.ListUsersByEmails(ctx, []string{email})
+
+	// User with e-mail already registered
+	if err == nil {
+		return PlatformInvite{}, dbutil.ErrConflict
+	} else if !errors.Contains(err, dbutil.ErrNotFound) {
```
bool3max: No need for an else-if here.

users/postgres/invites.go
```diff
@@ -0,0 +1,292 @@
...
+			tx.Rollback()
+
+			pgErr, ok := err.(*pgconn.PgError)
+			if ok {
+				switch pgErr.Code {
+				case pgerrcode.InvalidTextRepresentation:
+					return errors.Wrap(dbutil.ErrMalformedEntity, err)
+				case pgerrcode.UniqueViolation:
+					var e = dbutil.ErrConflict
+					if pgErr.ConstraintName == "ux_invites_platform_invitee_email" {
```
bool3max: No need for the constraint-name check, there's only one possible one.

users/postgres/init.go
```diff
@@ -77,6 +77,27 @@ func migrateDB(db *sqlx.DB) error {
...
+					CREATE TABLE IF NOT EXISTS invites_platform (
+						id            UUID NOT NULL,
+						invitee_email VARCHAR NOT NULL,
+						created_at    TIMESTAMPTZ,
+						expires_at    TIMESTAMPTZ,
+						state         VARCHAR DEFAULT 'pending' NOT NULL
+ 					)
+					`,
+					`
+					CREATE UNIQUE INDEX ux_invites_platform_invitee_email on invites_platform (invitee_email) WHERE state='pending'
```
bool3max: Rename index.

users/postgres/init.go
```diff
@@ -77,6 +77,27 @@ func migrateDB(db *sqlx.DB) error {
 					`DROP TABLE IF EXISTS verifications`,
 				},
 			},
+			{
+				Id: "users_3",
+				Up: []string{
+					`
+					CREATE TABLE IF NOT EXISTS invites_platform (
```
bool3max: Start string on line with backtick.

auth/api/http/invites/endpoint.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	return response
+}
+
+func buildOrgInviteRes(inv auth.OrgInvite) orgInviteRes {
+	return orgInviteRes{
+		ID:          inv.ID,
+		InviteeID:   inv.InviteeID,
+		InviteeRole: inv.InviteeRole,
```
majabirmancevic, line 134: would it be clear enough if we used only `Role`?

bool3max: I think so but I think there's no harm in InviteeRole either, it's not much longer but makes it immediately apparent to the reader what it's about.

auth/invites.go
```diff
@@ -0,0 +1,334 @@
...
+	ViewOrgInvite(ctx context.Context, token, inviteID string) (OrgInvite, error)
+
+	// ListOrgInvitesByUser retrieves a list of invites either directed towards a specific Invitee,
+	// or sent out by a specific Inviter, depending on the value of the `userType` argument, which
+	// must be either 'invitee' or 'inviter'.
+	ListOrgInvitesByUser(ctx context.Context, token, userType, userID string, pm PageMetadataInvites) (OrgInvitesPage, error)
+
+	// ListOrgInvitesByOrgID retrieves a list of invites towards any user(s) to join the org identified
+	// by its ID
+	ListOrgInvitesByOrgID(ctx context.Context, token, orgID string, pm PageMetadataInvites) (OrgInvitesPage, error)
```
majabirmancevic: `ListOrgInvitesByOrg`

majabirmancevic: hm, and it seems redundant.. what if we use `ListOrgInvites`?

auth/api/metrics.go
```diff
@@ -224,3 +224,66 @@ func (ms *metricsMiddleware) RetrieveRole(ctx context.Context, id string) (strin
 
 	return ms.svc.RetrieveRole(ctx, id)
 }
+
+func (ms *metricsMiddleware) CreateOrgInvite(ctx context.Context, token, email, role, orgID, invRedirectPath string) (auth.OrgInvite, error) {
+	defer func(begin time.Time) {
+		ms.counter.With("method", "invite_org_member").Add(1)
```
majabirmancevic: `create_org_invite`

auth/api/logging.go
```diff
@@ -313,3 +313,101 @@ func (lm *loggingMiddleware) RetrieveRole(ctx context.Context, id string) (role
 
 	return lm.svc.RetrieveRole(ctx, id)
 }
+
+func (lm *loggingMiddleware) CreateOrgInvite(ctx context.Context, token, email, role, orgID, invRedirectPath string) (invite auth.OrgInvite, err error) {
+	defer func(begin time.Time) {
+		message := fmt.Sprintf("Method invite_org_member took %s to complete", time.Since(begin))
```
majabirmancevic: `create_org_invite`

auth/invites.go
```diff
@@ -0,0 +1,334 @@
...
+
+	// RetrieveOrgInviteByID retrieves a specific OrgInvite by its ID.
+	RetrieveOrgInviteByID(ctx context.Context, inviteID string) (OrgInvite, error)
+
+	// RemoveOrgInvite removes a specific pending OrgInvite.
+	RemoveOrgInvite(ctx context.Context, inviteID string) error
+
+	// RetrieveOrgInviteByUserID retrieves a list of invites either directed towards a specific Invitee, or sent out by a
+	// specific Inviter, depending on the value of the `userType` argument, which must be either 'invitee' or 'inviter'.
+	RetrieveOrgInvitesByUserID(ctx context.Context, userType, userID string, pm PageMetadataInvites) (OrgInvitesPage, error)
```
majabirmancevic: `RetrieveOrgInvitesByUser`

auth/invites.go
```diff
@@ -0,0 +1,334 @@
...
+
+	// RetrieveOrgInvitesByOrgID retrieves a list of invites towards any user(s) to join the Org identified
+	// by its ID.
+	RetrieveOrgInvitesByOrgID(ctx context.Context, orgID string, pm PageMetadataInvites) (OrgInvitesPage, error)
```
majabirmancevic: `RetrieveOrgInvitesByOrg` (or `RetrieveOrgInvites`)

auth/invites.go
bool3max: Yeah the `ID` is def redundant. So I agree with `ListOrgInvitesByOrg`. 

As for just `ListOrgInvites`, well, I think it would be kind of confusing. In `ListOrgInvitesByOrg` the first `Org` denotes what type of invite we're talking about (Org invites vs Platform invites), while the the second one (in `ByOrg`) denotes based on what we're retrieving the invites, and so `ListOrgInvites` by itself would be ambiguous.

auth/invites.go
```diff
@@ -0,0 +1,334 @@
...
+
+	// UpdateOrgInviteState updates the state of a specific Invite denoted by its ID.
+	UpdateOrgInviteState(ctx context.Context, inviteID, state string) error
+}
+
+func (svc service) CreateOrgInvite(ctx context.Context, token, email, role, orgID, invRedirectPath string) (OrgInvite, error) {
+	// Check if currently authenticated User has "admin" or higher privileges within Org (required to make invitations)
```
majabirmancevic: we can omit a comment and emphasize it in the docs 

> (required to make invitations)

auth/invites.go
```diff
@@ -0,0 +1,334 @@
...
+	if err := svc.canAccessOrg(ctx, token, orgID, Admin); err != nil {
+		return OrgInvite{}, err
+	}
+
+	// Get userID of inviter
+	inviterIdentity, err := svc.identify(ctx, token)
```
majabirmancevic: `user`/`inviter` instead of `inviterIdentity`

and no need for comment

auth/invites.go
majabirmancevic: yes yes, I get it, but we have something similar for OrgMemberships in auth and GroupMemberships in things

the fact is that `ListOrgInvites` would make more sense for a list of all invites from all orgs, but we don't have that endpoint right now

auth/invites.go
```diff
@@ -0,0 +1,334 @@
...
+	if err != nil {
+		return OrgInvite{}, err
+	}
+
+	org, err := svc.ViewOrg(ctx, token, orgID)
```
majabirmancevic: can we directly use svc.orgs.RetrieveByID, since we previously checked canAccessOrg, and the same check exists in the service methods?

-> same for svc.memberships.Retrieve...

auth/invites.go
```diff
@@ -0,0 +1,334 @@
...
+		ExpiresAt:   createdAt.Add(svc.inviteDuration),
+		State:       InviteStatePending,
+	}
+
+	if err := svc.invites.SaveOrgInvite(ctx, invite); err != nil {
+		return OrgInvite{}, err
+	}
+
+	go func() {
+		svc.SendOrgInviteEmail(ctx, invite, email, org.Name, invRedirectPath)
```
majabirmancevic, line 170: maybe we could log an error

auth/invites.go
bool3max: Right!

auth/invites.go
bool3max: The error is logged by the logging middleware.

majabirmancevic: ah, right!

auth/invites.go
```diff
@@ -0,0 +1,334 @@
...
+	}()
+
+	return invite, nil
+}
+
+func (svc service) RevokeOrgInvite(ctx context.Context, token, inviteID string) error {
+	// Identify User attempting to revoke invite
+	currentUser, err := svc.identify(ctx, token)
```
majabirmancevic: `user`

auth/invites.go
```diff
@@ -0,0 +1,334 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Obtain full invite from db based on inviteID
```
majabirmancevic: no need for comment

(check other methods where the code is self-explanatory)

auth/invites.go
```diff
@@ -0,0 +1,334 @@
...
+	// the invite belongs
+	if err := svc.isAdmin(ctx, token); err == nil {
+		return invite, nil
+	}
+
+	if err := svc.canAccessOrg(ctx, token, invite.OrgID, Admin); err == nil {
+		return invite, nil
+	}
+
+	currentUser, err := svc.identify(ctx, token)
```
majabirmancevic: `user`

auth/invites.go
```diff
@@ -0,0 +1,334 @@
...
+	if currentUser.ID == invite.InviteeID {
+		return invite, nil
+	}
+
+	return OrgInvite{}, errors.ErrAuthorization
+}
+
+func (svc service) RespondOrgInvite(ctx context.Context, token, inviteID string, accept bool) error {
+	// Identify User attempting to respond to Invite
+	currentUser, err := svc.identify(ctx, token)
```
majabirmancevic: `user`

auth/invites.go
```diff
@@ -0,0 +1,334 @@
...
+
+	newState := InviteStateDeclined
+
+	if accept {
+		// User has accepted the Invite, assign them as a member of the appropriate Org
+		// with the appropriate role
+		newState = InviteStateAccepted
+		ts := getTimestmap()
+
+		newOrgMember := OrgMembership{
```
majabirmancevic: `membership`

auth/invites.go
```diff
@@ -0,0 +1,334 @@
...
+
+	return nil
+}
+
+func (svc service) ListOrgInvitesByOrgID(ctx context.Context, token, orgID string, pm PageMetadataInvites) (OrgInvitesPage, error) {
+	if err := svc.canAccessOrg(ctx, token, orgID, Admin); err != nil {
+		return OrgInvitesPage{}, err
+	}
+
+	invitesPage, err := svc.invites.RetrieveOrgInvitesByOrgID(ctx, orgID, pm)
```
majabirmancevic: or just `page`

auth/invites.go
```diff
@@ -0,0 +1,334 @@
...
+}
+
+func (svc service) ListOrgInvitesByUser(ctx context.Context, token, userType, userID string, pm PageMetadataInvites) (OrgInvitesPage, error) {
+	if err := svc.isAdmin(ctx, token); err != nil {
+		if err != errors.ErrAuthorization {
+			return OrgInvitesPage{}, err
+		}
+
+		// Current User is not Root Admin - must be either the Invitee or Inviter
+		currentUser, err := svc.identify(ctx, token)
```
majabirmancevic: `user`

auth/mocks/email.go
```diff
@@ -0,0 +1,18 @@
...
+
+type emailerMock struct {
+}
+
+// NewEmailer provides emailer instance for  the test
+func NewEmailer() auth.Emailer {
+	return &emailerMock{}
+}
+
+func (e *emailerMock) SendOrgInvite(To []string, inv auth.OrgInvite, orgName, invRedirectPath string) error {
```
majabirmancevic: `To` or `to`?

auth/mocks/invites.go
```diff
@@ -0,0 +1,162 @@
...
+func (irm *invitesRepositoryMock) SaveOrgInvite(ctx context.Context, invites ...auth.OrgInvite) error {
+	irm.mu.Lock()
+	defer irm.mu.Unlock()
+
+	for _, invite := range invites {
+		if _, ok := irm.orgInvites[invite.ID]; ok {
+			return dbutil.ErrConflict
+		}
+
+		for _, iInv := range irm.orgInvites {
```
majabirmancevic: `iInv` -> `oi`/`i` ?

auth/postgres/init.go
```diff
@@ -96,6 +96,28 @@ func migrateDB(db *sqlx.DB) error {
 				},
 				Down: []string{},
 			},
+			{
+				Id: "auth_3",
+				Up: []string{
+					`CREATE TABLE IF NOT EXISTS invites_org (
```
majabirmancevic: `org_invites`?

auth/service_test.go
```diff
@@ -1180,6 +1203,453 @@ func TestListOrgMemberships(t *testing.T) {
 	}
 }
 
+func TestInviteMembers(t *testing.T) {
```
majabirmancevic: `TestCreateOrgInvite`
and update test descriptions

auth/service_test.go
```diff
@@ -1180,6 +1203,453 @@ func TestListOrgMemberships(t *testing.T) {
...
+	}
+
+	for _, tc := range cases {
+		err := svc.RevokeOrgInvite(context.Background(), tc.token, testInviteID)
+		assert.True(t, errors.Contains(err, tc.err), fmt.Sprintf("%s expected %s got %s\n", tc.desc, tc.err, err))
+	}
+
+}
+
+func TestInviteRespond(t *testing.T) {
```
majabirmancevic: `TestRespondInvite`

auth/mocks/invites.go
bool3max: What do you think about `innerInvite`? I'm not a fan of the short oi/i

auth/tracing/invites.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"github.com/MainfluxLabs/mainflux/auth"
+	opentracing "github.com/opentracing/opentracing-go"
+)
+
+const (
+	saveOrgInvite              = "save_org_invite"
+	retrieveOrgInviteByID      = "retrieve_org_invite_by_id"
+	removeOrgInvite            = "remove_org_invite"
+	retrieveOrgInvitesByUserID = "retrieve_org_invites_by_user_id"
```
majabirmancevic: `retrieveOrgInvitesByUser`

cmd/auth/main.go
```diff
@@ -143,7 +170,7 @@ func main() {
 
 	tc := thingsapi.NewClient(thConn, thingsTracer, cfg.timeout)
 
-	svc := newService(db, tc, uc, dbTracer, cfg.secret, logger, cfg.loginDuration)
+	svc := newService(db, tc, uc, dbTracer, &cfg.emailConfig, cfg.secret, logger, cfg.loginDuration, cfg.inviteDuration, cfg.host)
```
majabirmancevic, line 173: maybe pass the whole `cfg` object instead of each field separately?

bool3max: Yeah. Maybe it'd be better to do it in a separate PR?

majabirmancevic: yeah, that's better

users/api/http/endpoint_test.go
```diff
@@ -1008,6 +1017,300 @@ func TestPasswordChange(t *testing.T) {
...
+
+	cases := []struct {
+		desc        string
+		req         string
+		contentType string
+		status      int
+		token       string
+	}{
+		{
+			"invite platform member",
```
majabirmancevic: "create platform invite"

users/api/http/transport.go
```diff
@@ -60,6 +62,13 @@ func MakeHandler(svc users.Service, tracer opentracing.Tracer, logger logger.Log
 		opts...,
 	))
 
+	mux.Post("/register/invite/:inviteID", kithttp.NewServer(
```
majabirmancevic: :id

users/api/http/transport.go
```diff
@@ -137,6 +146,34 @@ func MakeHandler(svc users.Service, tracer opentracing.Tracer, logger logger.Log
 		opts...,
 	))
 
+	mux.Post("/invites-platform", kithttp.NewServer(
```
majabirmancevic: `/invites` is enough?

users/invites.go
```diff
@@ -0,0 +1,200 @@
...
+
+type PageMetadataInvites struct {
+	apiutil.PageMetadata
+	State string `json:"state,omitempty"`
+}
+
+type PlatformInvites interface {
+	// InvitePlatformMember creates a pending platform Invite for the appropriate email address.
+	// Only usable by the platform Root Admin.
+	InvitePlatformMember(ctx context.Context, token, redirectPath, email string) (PlatformInvite, error)
```
majabirmancevic: `CreatePlatformInvite`

users/service.go
```diff
@@ -57,6 +57,12 @@ type Service interface {
 	// Returns the ID of the newly-registered User upon success.
 	VerifyEmail(ctx context.Context, confirmationToken string) (string, error)
 
+	// PlatformInviteRegister performs user registration based on a platform invite.
+	// inviteID must correspond to a valid, pending and non-expired platform invite, and the user's supplied
+	// e-mail address must match the e-mail address of that platform invite. Upon success, marks the associated
+	// invite's state as 'accepted'. Returns the ID of the newly registered user.
+	PlatformInviteRegister(ctx context.Context, user User, inviteID string) (string, error)
```
majabirmancevic: can we find a slightly clearer name for this? `RegisterByInvite`/`RegisterInvitedUser` or similar..

users/service.go
```diff
@@ -139,6 +147,8 @@ var _ Service = (*usersService)(nil)
 type usersService struct {
 	users               UserRepository
 	emailVerifications  EmailVerificationRepository
+	platformInvites     PlatformInvitesRepository
```
majabirmancevic: `invites`

users/service.go
```diff
@@ -225,6 +237,45 @@ func (svc usersService) SelfRegister(ctx context.Context, user User, redirectPat
...
+	_, err := svc.users.RetrieveByEmail(ctx, user.Email)
+	if err != nil && !errors.Contains(err, dbutil.ErrNotFound) {
+		return "", err
+	}
+
+	if err == nil {
+		return "", dbutil.ErrConflict
+	}
+
+	// Validate platform invite
```
majabirmancevic: no need for comment

auth/mocks/invites.go
majabirmancevic: hm, or saved/existing?

users/service.go
bool3max: `RegisterByInvite` sounds good

auth/mocks/invites.go
bool3max: `existing` sounds great

users/api/http/transport.go
bool3max: Hmm yes actually I think so. It's in a different service so there should be no confusion.

auth/api/http/invites/endpoint_test.go
```diff
@@ -0,0 +1,591 @@
...
+	mux := httpapi.MakeHandler(svc, mocktracer.New(), logger)
+	return httptest.NewServer(mux)
+}
+
+func toJSON(data interface{}) string {
+	jsonData, _ := json.Marshal(data)
+	return string(jsonData)
+}
+
+func TestInviteMembers(t *testing.T) {
```
majabirmancevic: TestCreateInvites

auth/api/http/invites/endpoint_test.go
```diff
@@ -0,0 +1,591 @@
...
+		},
+		{
+			desc:   "invite member with invalid request format",
+			req:    "{",
+			ct:     contentType,
+			token:  ownerToken,
+			status: http.StatusBadRequest,
+		},
+		{
+			desc:   "create org without content type",
```
majabirmancevic: fix this desc and others

auth/service.go
```diff
@@ -227,3 +235,11 @@ func (svc service) isAdmin(ctx context.Context, token string) error {
 
 	return nil
 }
+
+func ValidateRole(role string) error {
```
majabirmancevic: is it used somewhere else or only in requests.go?

bool3max: Not currently but there's I think a couple places where manual role checks could be replaced by this func, but I didn't want to do it in this PR.

majabirmancevic: aha ok, I just thought that helper might be in requests.go

bool3max: I'll check later but if the other checks are in requests.go then we can move the function there for sure

auth/api/metrics.go
```diff
@@ -224,3 +224,66 @@ func (ms *metricsMiddleware) RetrieveRole(ctx context.Context, id string) (strin
...
+		ms.counter.With("method", "list_org_invites_by_user").Add(1)
+		ms.latency.With("method", "list_org_invites_by_user").Observe(time.Since(begin).Seconds())
+	}(time.Now())
+
+	return ms.svc.ListOrgInvitesByUser(ctx, token, userType, userID, pm)
+}
+
+func (ms *metricsMiddleware) ListOrgInvitesByOrg(ctx context.Context, token, orgID string, pm auth.PageMetadataInvites) (auth.OrgInvitesPage, error) {
+	defer func(begin time.Time) {
+		ms.counter.With("method", "list_org_invites_by_org_id").Add(1)
```
majabirmancevic: list_org_invites_by_org

auth/invites.go
```diff
@@ -0,0 +1,332 @@
...
+	// SaveOrgInvite saves one or more pending org invites to the repository.
+	SaveOrgInvite(ctx context.Context, invites ...OrgInvite) error
+
+	// RetrieveOrgInviteByID retrieves a specific OrgInvite by its ID.
+	RetrieveOrgInviteByID(ctx context.Context, inviteID string) (OrgInvite, error)
+
+	// RemoveOrgInvite removes a specific pending OrgInvite.
+	RemoveOrgInvite(ctx context.Context, inviteID string) error
+
+	// RetrieveOrgInviteByUserID retrieves a list of invites either directed towards a specific Invitee, or sent out by a
```
majabirmancevic: RetrieveOrgInvitesByUser

auth/api/logging.go
```diff
@@ -313,3 +313,101 @@ func (lm *loggingMiddleware) RetrieveRole(ctx context.Context, id string) (role
...
+
+		lm.logger.Info(fmt.Sprintf("%s without errors.", message))
+	}(time.Now())
+
+	return lm.svc.ListOrgInvitesByUser(ctx, token, userType, userID, pm)
+}
+
+func (lm *loggingMiddleware) ListOrgInvitesByOrg(ctx context.Context, token string, orgID string, pm auth.PageMetadataInvites) (invPage auth.OrgInvitesPage, err error) {
+	defer func(begin time.Time) {
+		message := fmt.Sprintf("Method list_org_invites_by_org_id took %s to complete", time.Since(begin))
```
majabirmancevic: `list_org_invites_by_org`

auth/invites.go
```diff
@@ -0,0 +1,332 @@
...
+	UpdateOrgInviteState(ctx context.Context, inviteID, state string) error
+}
+
+func (svc service) CreateOrgInvite(ctx context.Context, token, email, role, orgID, invRedirectPath string) (OrgInvite, error) {
+	// Check if currently authenticated User has "admin" or higher privileges within Org
+	if err := svc.canAccessOrg(ctx, token, orgID, Admin); err != nil {
+		return OrgInvite{}, err
+	}
+
+	// Get userID of inviter
```
majabirmancevic: remove comments

auth/invites.go
```diff
@@ -0,0 +1,332 @@
...
+
+	go func() {
+		svc.SendOrgInviteEmail(ctx, invite, email, org.Name, invRedirectPath)
+	}()
+
+	return invite, nil
+}
+
+func (svc service) RevokeOrgInvite(ctx context.Context, token, inviteID string) error {
+	// Identify User attempting to revoke invite
```
majabirmancevic: remove comment

auth/invites.go
```diff
@@ -0,0 +1,332 @@
...
+	return nil
+}
+
+func (svc service) ViewOrgInvite(ctx context.Context, token, inviteID string) (OrgInvite, error) {
+	invite, err := svc.invites.RetrieveOrgInviteByID(ctx, inviteID)
+	if err != nil {
+		return OrgInvite{}, err
+	}
+
+	// A specific Invite can only be retrieved by the platform Root Admin, the Invitee towards who
```
majabirmancevic: we will put the details in the docs

auth/invites.go
```diff
@@ -0,0 +1,332 @@
...
+	return OrgInvite{}, errors.ErrAuthorization
+}
+
+func (svc service) RespondOrgInvite(ctx context.Context, token, inviteID string, accept bool) error {
+	user, err := svc.identify(ctx, token)
+	if err != nil {
+		return err
+	}
+
+	// Obtain detailed information about the Invite
```
majabirmancevic: no need for comment

auth/postgres/init.go
```diff
@@ -96,6 +96,28 @@ func migrateDB(db *sqlx.DB) error {
...
+						invitee_role VARCHAR(12) NOT NULL,
+						created_at   TIMESTAMPTZ,
+						expires_at   TIMESTAMPTZ,
+						state        VARCHAR DEFAULT 'pending' NOT NULL,      
+						FOREIGN KEY  (org_id) REFERENCES orgs (id) ON DELETE CASCADE,
+						PRIMARY KEY  (id)
+					)`,
+					`CREATE UNIQUE INDEX unique_org_invitee_pending on org_invites (invitee_id, org_id) WHERE state='pending'`,
+				},
+				Down: []string{
```
majabirmancevic, line 116: not sure if we need `down` @manuio

auth/postgres/invites.go
```diff
@@ -0,0 +1,423 @@
...
+		return errors.Wrap(dbutil.ErrRemoveEntity, err)
+	}
+
+	return nil
+}
+
+func (ir invitesRepository) UpdateOrgInviteState(ctx context.Context, inviteID, state string) error {
+	query := `
+		UPDATE org_invites
+		SET state=:state
```
majabirmancevic: maybe add a space before `:state`?

and I'm not sure if the definition of query strings is consistent with other services?

auth/postgres/invites.go
```diff
@@ -0,0 +1,423 @@
...
+
+// Syncs the state of the Invite with the passed inviteID. That is, sets state='expired' if state='pending' and expires_at < now().
+func (ir invitesRepository) syncOrgInviteStateByID(ctx context.Context, inviteID string) error {
+	query := `
+		UPDATE org_invites
+		SET state='expired'
+		WHERE id=:inviteID AND state='pending' AND expires_at < NOW()
+	`
+
+	_, err := ir.db.NamedExecContext(ctx, query, map[string]any{"inviteID": inviteID})
```
majabirmancevic, line 373: we could probably use a helper for this repetitive part of the code (`syncOrgInviteStateByUserID`,`syncOrgInviteStateByInvite`)

users/api/http/endpoint.go
```diff
@@ -26,6 +26,22 @@ func selfRegistrationEndpoint(svc users.Service) endpoint.Endpoint {
 	}
 }
 
+func platformInviteRegistrationEndpoint(svc users.Service) endpoint.Endpoint {
```
majabirmancevic: registerByInviteEndpoint?

users/api/http/requests.go
```diff
@@ -43,6 +44,19 @@ func (req selfRegisterUserReq) validate() error {
 	return req.User.Validate(userPasswordRegex)
 }
 
+type platformInviteRegisterUserReq struct {
```
majabirmancevic: registerByInviteReq

users/api/http/responses.go
```diff
@@ -179,6 +180,72 @@ func (res deleteRes) Empty() bool {
...
+
+func (res createdPlatformInviteRes) Code() int {
+	if res.created {
+		return http.StatusCreated
+	}
+
+	return http.StatusOK
+}
+
+func (res createdPlatformInviteRes) Headers() map[string]string {
```
majabirmancevic: createPlatformInviteRes

auth/postgres/invites.go
bool3max: I've debated merging all three of these helper methods into one but eventually decided against it as it would complicate the signatures of each of the functions (more arguments and more logic inside). This way I think it's clearer exactly what each one does, and really the only duplicated part of code is the the one which executes the DB query and checks for errors.

auth/postgres/invites.go
bool3max: Yup I'll add spacing around `:state`. But I think the formatting here is more readable than elsewhere in the code where db queries are defined.

users/api/http/endpoint.go
bool3max: Went with this because the self-register endpoint is named `selfRegistrationEndpoint`. So maybe `inviteRegistrationEndpoint`?

auth/postgres/invites.go
majabirmancevic: yes it's okay that the helpers are separate, I thought we should put this code as a helper, but it's not necessary
```
_, err := ir.db.NamedExecContext(ctx, query, map[string]any{"inviteID": inviteID})
	if err != nil {
		pqErr, ok := err.(*pgconn.PgError)
		if ok {
			switch pqErr.Code {
			case pgerrcode.InvalidTextRepresentation:
				return errors.Wrap(dbutil.ErrMalformedEntity, err)
			}
		}
		return errors.Wrap(dbutil.ErrUpdateEntity, err)
	}
```

auth/postgres/invites.go
majabirmancevic: probably yes, we should improve then everywhere

users/api/http/endpoint.go
majabirmancevic: or selfRegistration should be consistent with other names, but we will consider that later;

in any case, better `inviteRegistrationEndpoint`

auth/postgres/init.go
bool3max: I don't believe we need an explicit `down` for the index, dropping the table will drop the associated index automatically, but I included it for clarity because the index is created using a separate SQL statement in the associated UP migration. (it's a "partial index" so it can't be embedded in the CREATE TABLE statement)

auth/invites.go
```diff
@@ -0,0 +1,326 @@
...
+		return errors.ErrAuthorization
+	}
+
+	newState := InviteStateDeclined
+
+	if accept {
+		// User has accepted the Invite, assign them as a member of the appropriate Org
+		// with the appropriate role
+		newState = InviteStateAccepted
+		ts := getTimestmap()
```
anna02272, line 263: This is typo right? But it's now shared across all services
 `getTimestamp()`

auth/invites.go
```diff
@@ -0,0 +1,326 @@
...
+	if err != nil {
+		return err
+	}
+
+	// An Invite can only be revoked by its issuer
+	if invite.InviterID != user.ID {
+		return errors.ErrAuthorization
+	}
+
+	if invite.State != InviteStatePending {
```
anna02272, line 192: Maybe this could be small helper function? In both RevokeOrgInvite and RespondOrgInvite

auth/invites.go
bool3max: Seems so, I never noticed that :D we can fix it in a NOISSUE PR separately.

auth/invites.go
bool3max: Hmm I'm not sure, it's only two places and the check is very basic, maybe it's a bit overkill for an entire function? I may be wrong though..

manuio: It's ok like that. Let's keep it

manuio (approved):
LGTM

majabirmancevic (approved):

PR #756, opened 2025-07-18
MF-757 - Add limit/offset to grpc method for retrieving users

### LINKED ISSUE #757: Add limit/offset to grpc method for retrieving users
Currently grpc method GetUsersByIDs does not include limit/offset data, even though the actual method supports it.
Include those parameters in the method and change the ListOrgMemberships and ListGroupMemberships methods to include the metadata returned by the GetUsersByIDs method.

Nesa002:
Limit/offset and total added to GetUsersByIDs (real and mock methods).
Fixed bugs in ListOrgMemberships and ListGroupMemberships where some memberships could be omitted due to improper pagination.
Note: Pagination and filtering currently applies to users fetched, not directly on memberships, which may cause edge cases and should be addressed in future improvements.
Resolves #757


majabirmancevic (approved):

manuio (approved):
LGTM

Issue #761, opened 2025-07-22
Change backup methods to return files and add restore endpoints


Nesa002:
Currently, backup methods return JSON body. Change them to return JSON files.
Add restore endpoints for each backup endpoint

PR #762, opened 2025-07-22
MF-761 – Add file-based backup and restore

Nesa002:
References #761


auth/api/http/memberships/transport.go
```diff
@@ -182,6 +213,29 @@ func encodeResponse(_ context.Context, w http.ResponseWriter, response interface
...
 
+func encodeFileResponse(_ context.Context, w http.ResponseWriter, response interface{}) (err error) {
+	w.Header().Set("Content-Type", apiutil.ContentTypeOctetStream)
+
+	if fr, ok := response.(viewFileRes); ok {
+		for k, v := range fr.Headers() {
+			w.Header().Set(k, v)
+		}
+
+		w.Header().Set("Content-Disposition", "attachment")
```
anna02272: Maybe we could consider adding a filename in the header here to suggest a default download name? @manuio

auth/api/http/memberships/transport.go
```diff
@@ -36,7 +38,14 @@ func MakeHandler(svc auth.Service, mux *bone.Mux, tracer opentracing.Tracer, log
 
 	mux.Get("/orgs/:id/memberships/backup", kithttp.NewServer(
 		kitot.TraceServer(tracer, "backup_org_memberships")(backupOrgMembershipsEndpoint(svc)),
-		decodeBackup,
+		decodeBackupByOrg,
+		encodeFileResponse,
+		opts...,
+	))
+
+	mux.Post("/orgs/:id/memberships/backup", kithttp.NewServer(
```
anna02272: Wouldn't it make more sense to use` /restore` as the route here, since the operation being performed is a restore, not a backup?

auth/api/http/memberships/transport.go
manuio: I agree

auth/api/http/memberships/transport.go
manuio: I thought that having GET and POST /backup is clear enough

anna02272: Then we should rename our /restore endpoint elsewhere

manuio: After to think about having different names is better. Let's keep /backup and /restore

auth/api/http/memberships/endpoint.go
```diff
@@ -102,15 +104,34 @@ func removeOrgMembershipsEndpoint(svc auth.Service) endpoint.Endpoint {
 
 func backupOrgMembershipsEndpoint(svc auth.Service) endpoint.Endpoint {
 	return func(ctx context.Context, request interface{}) (interface{}, error) {
-		req := request.(backupReq)
+		req := request.(backupByOrgReq)
 		if err := req.validate(); err != nil {
 			return nil, err
 		}
```
majabirmancevic, line 110: add blank line

things/api/http/groups/endpoint_test.go
```diff
@@ -26,31 +26,34 @@ import (
...
-	descKey          = "desc"
-	validData        = `{"limit":5,"offset":0}`
-	descData         = `{"limit":5,"offset":0,"dir":"desc","order":"name"}`
-	ascData          = `{"limit":5,"offset":0,"dir":"asc","order":"name"}`
-	invalidOrderData = `{"limit":5,"offset":0,"dir":"asc","order":"wrong"}`
-	zeroLimitData    = `{"limit":0,"offset":0}`
-	invalidDirData   = `{"limit":5,"offset":0,"dir":"wrong"}`
-	invalidLimitData = `{"limit":210,"offset":0}`
-	invalidData      = `{"limit": "invalid"}`
+	contentTypeJson        = "application/json"
```
majabirmancevic: contentTypeJSON

auth/api/http/memberships/transport.go
```diff
@@ -182,6 +214,28 @@ func encodeResponse(_ context.Context, w http.ResponseWriter, response interface
 	return json.NewEncoder(w).Encode(response)
 }
 
+func encodeFileResponse(_ context.Context, w http.ResponseWriter, response interface{}) (err error) {
+	w.Header().Set("Content-Type", apiutil.ContentTypeOctetStream)
```
majabirmancevic: can we move `encodeFileResponse` function to pkg/apiutil/transport.go and use it in all services, to reduce code duplication?

things/api/http/memberships/endpoint_test.go
```diff
@@ -26,24 +26,27 @@ import (
...
-	wrongValue     = "wrong_value"
-	password       = "password"
-	orgID          = "374106f7-030e-4881-8ab0-151195c29f92"
-	n              = 5
-	emailKey       = "email"
-	idKey          = "id"
-	ascKey         = "asc"
-	descKey        = "desc"
+	emptyValue             = ""
+	contentTypeJson        = "application/json"
```
majabirmancevic: contentTypeJSON

auth/api/http/memberships/transport.go
Nesa002: I moved it to apiutil along with the viewFileRes. If we don't want these changes I can revert the last commit. Probably also encodeResponse should be moved then if it is also the same everywhere, but that should probably go into a different issue since this pr has enough responsibilities

anna02272 (approved):

bool3max (approved):

manuio (approved):
LGTM

PR #772, opened 2025-08-01
MF-771 - Remove host from password reset request body

### LINKED ISSUE #771: Use Referer header instead of request body for password reset host field

anna02272:
Resolves #771


users/api/http/requests.go
```diff
@@ -117,18 +117,14 @@ func (req updateUserReq) validate() error {
 
 type passwResetReq struct {
 	Email string `json:"email"`
-	Host  string `json:"host"`
+	Host  string
```
majabirmancevic: should we remove the Host field from the request or just its validation?

anna02272: We should remove it entirely, since we're extracting the host from the Referer header.

majabirmancevic: ok, I was a bit confused by the PR title

users/api/http/transport.go
```diff
@@ -323,7 +323,7 @@ func decodePasswordResetRequest(_ context.Context, r *http.Request) (interface{}
 		return nil, errors.Wrap(apiutil.ErrMalformedEntity, err)
 	}
 
-	req.Host = r.Header.Get("Referer")
+	req.Host = r.Host
```
majabirmancevic: does the `Host` field need to be exported?

bool3max (approved):
LGTM

manuio (approved):
LGTM

PR #775, opened 2025-08-04
NOISSUE - Idiomatic return in users:emailer.New

bool3max:
Previously, if obtaining an emailer Agent failed, we returned a partially-initialized instance of `emailer` and a propagated error from `email.New(...)`. I thought this was fine because, from the caller's point of view the non-nil error signifies that the return value is not valid, but the most idiomatic (and safest) thing to do here is to return `nil` instead.


manuio (approved):
LGTM

duolok:
LGTM

duolok (approved):

PR #776, opened 2025-08-04
MF-777 - Refactor building URLs in user-facing e-mails

### LINKED ISSUE #777: Refactor the way URLs to the frontend in user-facing e-mails are constructed
Currently, the *host* of the URLs is supplied by the client (using the value of the HTTP `Referer` header) and blindly trusted by the backend, which poses a security risk.

bool3max:
This PR refactors the way URLs to the UI frontend are constructed in user-facing e-mails, such as those sent during e-mail verification and password reset flows.

Consider the following URL that a User may receive when trying to reset their password:

```
To reset your password, navigate to the following URL:

https://mainflux.net/auth/password-reset?token=XXX
        ^            ^
        |            |
       HOST         PATH
```

## Changes

1. The *HOST* of the UI platform is now set at service runtime, using the `MF_HOST` environment variable.
    * Previously, the host portion of the URL was supplied by the caller, i.e. API consumer, which posed a security risk.
2. The *PATH* on the UI platform to which the user will be redirected to when navigating to the URL from the e-mail is now supplied **in the request itself** - more specifically in the JSON object in the body of the request. 
    * Requests to which this applies, i.e. requests that may result in the user receiving an e-mail message, now also take a `"redirect_path"` parameter in the JSON object of the request's body.
    * Currently, the only two API endpoints to which this applies are:
    1. **The self-register endpoint (because of e-mail verification)**
    2. **The password reset request endpoint**

For example, the structure of the JSON object for the self-register endpoint (`/register`) now looks like so:

```json
{
    "user": {
        "email": "user@email.com",
        "password": "12345678"
    },
    "redirect_path": "/auth/email-verify"
}
```

**QUESTION**: Should we flatten the structure of the JSON object? In other words, should it look like this instead:

```json
{
    "email": "user@email.com",
    "password": "12345678"
    "redirect_path": "/auth/email-verify"
}
```


cmd/users/main.go
```diff
@@ -130,6 +131,7 @@ type config struct {
 	adminEmail          string
 	adminPassword       string
 	passRegex           *regexp.Regexp
+	host                string
 	selfRegisterEnabled bool
 	emailVerifyEnabled  bool
```
manuio, line 136: Remove the unused fields

manuio (changes requested):

cmd/users/main.go
bool3max: `emailVerifyEnabled` is used. 

`selfRegisterEnabled` has existed for a longer time (under a different name until recently), but its value was never actually used in the service in any way, so I guess we can remove it? @manuio

pkg/sdk/go/users.go
```diff
@@ -47,7 +49,14 @@ func (sdk mfSDK) CreateUser(u User, token string) (string, error) {
 }
 
 func (sdk mfSDK) RegisterUser(u User) (string, error) {
-	data, err := json.Marshal(u)
+	data, err := json.Marshal(struct {
+		User         User
+		RedirectPath string `json:"redirectPath"`
```
manuio: `redirect_path`

manuio (changes requested):

cmd/users/main.go
manuio: Yes, remove it please.

manuio: Wait, not in this PR

pkg/apiutil/errors.go
```diff
@@ -108,8 +108,8 @@ var (
 	// ErrMissingEmailToken indicates a missing e-mail verification token.
 	ErrMissingEmailToken = errors.New("missing e-mail verification token")
 
-	// ErrMissingHost indicates missing host.
-	ErrMissingHost = errors.New("missing host")
+	// ErrMissingPath indicates missing endpoint path from the frontend.
+	ErrMissingPath = errors.New("missing path")
```
manuio: `	ErrMissingRedirectPath = errors.New("missing redirect path")`

manuio (changes requested):

users/mocks/email.go
```diff
@@ -19,6 +19,6 @@ func (e *emailerMock) SendPasswordReset([]string, string, string) error {
 	return nil
 }
 
-func (e *emailerMock) SendEmailVerification(To []string, host, token string) error {
+func (e *emailerMock) SendEmailVerification(To []string, path, token string) error {
```
anna02272: Should it be `redirectPath` here as well?

bool3max: Yes! Good catch

anna02272 (approved):

manuio (approved):
LGTM

PR #784, opened 2025-08-12
MF-703 -  Support multiple conditions per rule

### LINKED ISSUE #703: Support multiple conditions per rule
Currently, it is possible to define a single condition when creating a rule. It is necessary to enable multiple conditions per rule, so we can support more complex checks, for example `field1 >= val1 && field2 <= val2`.

majabirmancevic:
Resolves #703

### DIFF
--- a/rules/api/http/requests.go
+++ b/rules/api/http/requests.go
@@ -5,6 +5,7 @@ package http
 
 import (
 	"github.com/MainfluxLabs/mainflux/pkg/apiutil"
+	"github.com/MainfluxLabs/mainflux/pkg/errors"
 	"github.com/MainfluxLabs/mainflux/rules"
 )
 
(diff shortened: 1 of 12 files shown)


anna02272 (dismissed):

rules/service.go
```diff
@@ -233,75 +238,105 @@ func (rs *rulesService) Publish(ctx context.Context, message protomfx.Message) e
...
+		case float64:
+			payloadValue = v
+		case int:
+			payloadValue = float64(v)
+		case int64:
+			payloadValue = float64(v)
+		case uint:
+			payloadValue = float64(v)
+		case uint64:
+			payloadValue = float64(v)
```
manuio, line 298: inline these cases

rules/service.go
```diff
@@ -46,6 +46,12 @@ const (
 	ActionTypeSMPP  = "smpp"
 	ActionTypeAlarm = "alarm"
 )
+
+const (
+	OperatorAND = "AND"
+	OperatorOR  = "OR"
+)
+
 const subjectAlarm = "alarms"
```
manuio: Why there are 3 const blocks?

manuio (changes requested):

rules/api/http/requests.go
```diff
@@ -14,12 +15,33 @@ const (
...
+	errMissingConditionThreshold = errors.New("missing condition threshold")
+
+	// errInvalidActionType indicates an invalid action type
+	errInvalidActionType = errors.New("missing or invalid action type")
+
+	// errMissingActionID indicates a missing action id
+	errMissingActionID = errors.New("missing action id")
+
+	// errInvalidOperator indicates an invalid logical operator
+	errInvalidOperator = errors.New("missing or invalid logical operator")
```
manuio: These errors are not going to be loggeg here: https://github.com/MainfluxLabs/mainflux/blob/master/pkg/apiutil/transport.go#L56

manuio (changes requested):

manuio (approved):
LGTM

rules/service.go
bool3max: Is this resolved?

anna02272 (approved):

rules/service.go
majabirmancevic: cannot be inlined bcs of interface conversion to float64 `float64(v)`

bool3max: oh, right!

bool3max (approved):

PR #793, opened 2025-08-14
MF-790 - Asynchronously send e-mails from service methods

### LINKED ISSUE #790: Asynchronously send e-mails from service methods
Service methods which send e-mails (such as those for self-registration or password reset) currently do so synchronously, blocking the incoming HTTP/gRPC request that initiated them and thus making the user wait more than is necessary.

We should try and send the e-mails asynchronously, probably in a separate goroutine.

bool3max:
Addresses #790 by calling e-mail related methods from goroutines from within service methods.

However, I've got a question. As we're no longer propagating errors from e-mail related methods *outside* of the service methods that call them (because they're now called from goroutines), the HTTP logging middleware doesn't pick up on potential errors and sending emails can fail "silently" -- see for example [here](https://github.com/bool3max/mainflux/blob/8fca1883a98c702555f5e276b91013b12e8f08be/users/service.go#L501): the `error` return value from `svc.SendPasswordReset` never makes it out of `GenerateResetToken`.

One potential solution is to employ a "logging middleware", similar to that used by HTTP API handlers and services, but for the [`Emailer`](https://github.com/bool3max/mainflux/blob/8fca1883a98c702555f5e276b91013b12e8f08be/users/emailer.go#L7). 

Another slightly simpler one is to make the [concrete implementation of `Emailer`](https://github.com/bool3max/mainflux/blob/8fca1883a98c702555f5e276b91013b12e8f08be/users/emailer/emailer.go#L20) accept a `logger` and let it log errors from its methods, itself.

Thoughts?

### DIFF
--- a/cmd/users/main.go
+++ b/cmd/users/main.go
@@ -292,14 +292,31 @@ func newService(db *sqlx.DB, tracer opentracing.Tracer, ac protomfx.AuthServiceC
 	userRepo := tracing.UserRepositoryMiddleware(postgres.NewUserRepo(database), tracer)
 	verificationRepo := tracing.VerificationRepositoryMiddleware(postgres.NewEmailVerificationRepo(database), tracer)
 
-	emailer, err := emailer.New(c.host, &c.emailConf)
+	svcEmailer, err := emailer.New(c.host, &c.emailConf)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


bool3max:
The problem has been solved by implementing logging & metrics middlewares for the `users.Emailer` implementation.

manuio (approved):
LGTM

PR #794, opened 2025-08-14
MF-703 - Update rules_2 migration

majabirmancevic:
Migration fix: set default for operator and convert legacy condition object to array with comparator

Related to #703


manuio (approved):
LGTM

PR #798, opened 2025-08-20
NOISSUE - Readers SQL query fail

duolok:


manuio (dismissed):
LGTM

readers/postgres/messages.go
```diff
@@ -301,7 +301,7 @@ func (tr postgresRepository) fmtCondition(rpm readers.PageMetadata, table string
 			default:
 				condition = fmt.Sprintf(`%s %s name = :name`, condition, op)
 			}
-
+			op = "AND"
```
majabirmancevic: we can avoid repeating `op = "AND"`, so here's an example of refactoring:
```
condition := ""
op := "WHERE"

for name := range query {
    var clause string
    switch name {
    case "subtopic", "publisher", "protocol":
        clause = fmt.Sprintf("%s = :%s", name, name)
    case "name":
        switch table {  
        case jsonTable:
            clause = "payload->>'n' = :name"
        default:
            clause = "name = :name"
        }
    case "v":
        comparator := readers.ParseValueComparator(query)
        clause = fmt.Sprintf("value %s :value", comparator)
    case "vb":
        clause = "bool_value = :bool_value"
    case "vs":
        clause = "string_value = :string_value"
    case "vd":
        clause = "data_value = :data_value"
    case "from":
        clause = fmt.Sprintf("%s >= :from", timeColumn)
    case "to":
        clause = fmt.Sprintf("%s <= :to", timeColumn)
    default:
        continue
    }

    condition = fmt.Sprintf("%s %s %s", condition, op, clause)
    op = "AND"
```

manuio (approved):
LGTM

majabirmancevic (approved):

PR #806, opened 2025-08-25
MF-803 - Fix permission check in svc.ViewUser

### LINKED ISSUE #803: Missing permission/identity check in users.ViewUser
[`users.ViewUser`](https://github.com/MainfluxLabs/mainflux/blob/8e7e409541aa15216debbef5ab475a8f094c9aeb/users/service.go#L354) makes a call to `identify()` but it ignores the result, making any user able to retrieve the profile of any other user.

bool3max:
Simple. Fixes #803. To view a specific user's profile, the user making the request must be authenticated as either the Root Admin or as the user whose profile they're trying to view.

### DIFF
--- a/users/service.go
+++ b/users/service.go
@@ -352,8 +352,15 @@ func (svc usersService) Login(ctx context.Context, user User) (string, error) {
 }
 
 func (svc usersService) ViewUser(ctx context.Context, token, id string) (User, error) {
-	if _, err := svc.identify(ctx, token); err != nil {
+	currentUser, err := svc.identify(ctx, token)
+	if err != nil {
 		return User{}, err
+	}
+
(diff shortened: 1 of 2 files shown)


users/service.go
```diff
@@ -352,10 +352,17 @@ func (svc usersService) Login(ctx context.Context, user User) (string, error) {
 }
 
 func (svc usersService) ViewUser(ctx context.Context, token, id string) (User, error) {
-	if _, err := svc.identify(ctx, token); err != nil {
+	currentUser, err := svc.identify(ctx, token)
```
manuio: Rename it `user`

manuio (changes requested):

majabirmancevic (approved):

manuio (approved):
LGTM

PR #807, opened 2025-08-25
MF-802 - Disable self-register at runtime in users service

### LINKED ISSUE #802: Make use of selfRegisterEnabled flag in users service
We parse the `MF_USERS_SELF_REGISTER_ENABLED` env. var. and [load it into the `config`](https://github.com/MainfluxLabs/mainflux/blob/8e7e409541aa15216debbef5ab475a8f094c9aeb/cmd/users/main.go#L135), but we never pass it to the service, making it effectively useless.

We should pass it when initializing the service, and then use its value to enable/disable self-register.

The question is how do we disable self-registration:

1. If `selfRegisterEnabled` is false, skip creating the self-register endpoint in [`MakeHandler`](https://github.com/MainfluxLabs/mainflux/blob/8e7e409541aa15216debbef5ab475a8f094c9aeb/users/api/http/transport.go#L49)
or
2. Keep the endpoint registered, but check the `selfRegisterEnabled` value in every request and abort as appropriate

bool3max:
Fixes #802 by passing the `selfRegisterEnabled` config flag to the users service, and checking for its value in the `SelfRegister` service method, in which case it returns `ErrSelfRegisterDisabled`, which gets encoded as a `501 Not Implemented` HTTP status code.

### DIFF
--- a/cmd/users/main.go
+++ b/cmd/users/main.go
@@ -316,7 +316,7 @@ func newService(db *sqlx.DB, tracer opentracing.Tracer, ac protomfx.AuthServiceC
 
 	idProvider := uuid.New()
 
-	svc := users.New(userRepo, verificationRepo, c.emailVerifyEnabled, hasher, ac, svcEmailer, idProvider)
+	svc := users.New(userRepo, verificationRepo, c.emailVerifyEnabled, c.selfRegisterEnabled, hasher, ac, svcEmailer, idProvider)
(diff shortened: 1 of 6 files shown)


manuio (approved):
LGTM

majabirmancevic (approved):

duolok (approved):

PR #822, opened 2025-09-02
MF-821 - Check for apiutil, dbutil, and "errors" package errors in apiutil.EncodeError only

### LINKED ISSUE #821: Refactor apiutil.EncodeError
`apiutil.EncodeError` should check for all possible `pkg/apiutil` and `pkg/dbutil` errors and encode them to the response header appropriately, while `encodeError` functions specific to each service's HTTP transport layer should only check for local, service-specific errors.

bool3max:
When encoding error values to HTTP response status codes, now `apiutil.EncodeError` checks for all `apiutil`/`dbutil`/`pkg/errors` errors, while local `encodeError()` functions in each service's transport layer only check for errors specific to that service.

Additionally some unused errors were removed.


manuio (approved):
LGTM

majabirmancevic (approved):

PR #830, opened 2025-09-08
MF-829 - Use separator for nested Condition.Field

### LINKED ISSUE #829: Use separator for nested Condition.Field
Currently, we use a recursive parameter search through the message payload to check the condition of the rule.
It can happen that within the payload several JSON objects contain a field with the same name, so we can get an unexpected result, for example:
If we have a payload:

```
{
    "obj1": {
        "param1": 15,
        "param2": 40,
        "current": 30
    },
    "obj2": {
        "param3": 10,
        "param4": 100,
        "current": 50
    }
}
```

and if we define a Condition with parameters:` "field":"current", "comparator":">=", "threshold":30`,
an alarm/notification can be created for a wrong parameter. 
Instead, the definition of Condition.Field should be `obj1.current` or `obj2.current`, and thus avoid confusion.

majabirmancevic:
Resolves #829

### DIFF
--- a/pkg/messaging/pubsub.go
+++ b/pkg/messaging/pubsub.go
@@ -166,22 +166,6 @@ func FormatMessage(pc *protomfx.PubConfByKeyRes, msg *protomfx.Message) error {
 	return nil
 }
 
-func FindParam(payload map[string]interface{}, param string) interface{} {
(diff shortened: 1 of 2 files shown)


manuio (approved):
LGTM

Issue #773, opened 2025-08-01
Not all relevant/applied page filters are returned in the response.


BraceZenith:
When we log the response from `this.groupsService.getThingsetThings()` for example, we get the attached screenshot.
Now the thing is, I am on the 3rd page (offset 20), showing 10 items per page (limit 10), but I'm also sorting by ID, in an ascending order, and that part is not reflected anywhere in the response, even though those filters are just as important in terms of what we are getting.
Is it possible to update the response, to feature all the relevant filters?
I'd need it for Table component for example, so that I can respond with the Table to possible preset configurations regarding sorting, and not just offset and limit.
If user refreshes the page, even though the sorting is respected, due to URL query params, it is not reflected in the UI element's styles/appearances, because the Table component's `data: Page` `@Input`  doesn't hold `order` and `direction` properties. Only `offset` `limit` and `total`. So I can't use that information, to make the table properly indicate the selected column and the sorting direction.

<img width="618" height="121" alt="Image" src="https://github.com/user-attachments/assets/f8ea0dab-2d99-4dba-84dc-5b5a61d1fa3e" />

PR #844, opened 2025-09-18
MF-843 - Refactor return value of service methods for retreving mulitple resources

### LINKED ISSUE #843: Refactor service methods for retrieving multiple resources not to return PageMetadata
Service methods across the entire codebase whose purpose is to retrieve a list of resources based on "PageMetadata" filters (i.e. offset, limit, resource name, etc..) almost always return the passed `PageMetadata` back to the caller.

However, in almost all cases the returned `PageMetadata` information is exactly the same as the information passed in by the caller, making it useless. The only exception is the "`Total`" field, which the caller doesn't supply in the first place, and which the service method must return back.

Service methods that follow this pattern should be refactored to return only a slice of the appropriate resource alongside the `Total` count of the resource in the database.

bool3max:
Closes #843. 

Also partially addresses #773. I say partially because that issue is present for different reasons in different services. In services such as `things` for example, where the issue was due to PageMetadata information being lost somewhere along the way from service method -> repository method call (because the repository method wasn't returning back all relevant PageMetadata fields), it is fixed by this PR - because from HTTP endpoints we now return an exact copy of query metadata that was supplied in the HTTP request.

In some other services (such as `users`), that issue is present because HTTP response structs don't encode all relevant PageMetadata fields - i.e. [here](https://github.com/MainfluxLabs/mainflux/blob/d966b6c5597db6636bd9fc6371f450299fc3faf0/users/api/http/responses.go#L25), and so it will have to be addressed separately.

### DIFF
--- a/auth/api/http/invites/endpoint.go
+++ b/auth/api/http/invites/endpoint.go
@@ -86,7 +86,7 @@ func listOrgInvitesByUserEndpoint(svc auth.Service, userType string) endpoint.En
 			return nil, err
 		}
 
-		response := buildOrgInvitesPageRes(page)
+		response := buildOrgInvitesPageRes(page, req.pm)
 
 		return response, nil
 	}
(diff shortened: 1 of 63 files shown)


manuio (approved):
LGTM

majabirmancevic (approved):

PR #846, opened 2025-09-19
MF-773 - Return all relevant PageMetadata filters in HTTP API responses

### LINKED ISSUE #773: Not all relevant/applied page filters are returned in the response.

bool3max:
Closes #773. Should be fixed in all services.


duolok (approved):
LGTM

manuio (approved):
LGTM

PR #856, opened 2025-09-29
MF-862 - Readers publisher endpoint

### LINKED ISSUE #862: Add publisher endpoint for readers
Right now, we only support listing messages using Thing key, enhance readers to support getting messages with user token and add publisher filter.

duolok:
Resolves #862


manuio (approved):
LGTM

anna02272 (approved):

PR #865, opened 2025-10-03
MF- 864 - Add access check for publisher in messages listing and backup

### LINKED ISSUE #864: Add proper access check for publisher in messages endpoints

anna02272:
Resolves #864


readers/service.go
```diff
@@ -15,6 +15,8 @@ const (
 	senmlFormat = "senml"
 	csvFormat   = "csv"
 	rootSubject = "root"
+
+	Viewer = "viewer"
```
manuio: Use the const from auth service instead of duplicating this one

manuio (changes requested):

readers/service.go
anna02272: I'm getting `import cycle not allowed` error if I use it from auth or things

majabirmancevic: > Use the const from auth service instead of duplicating this one

we should move consts to pkg because we also have them in [things](https://github.com/MainfluxLabs/mainflux/blob/2a37193c5b28eae31b9cd887f4e10112b2ccd51a/things/service.go#L18)

majabirmancevic (approved):

duolok (approved):

manuio (approved):
LGTM

PR #867, opened 2025-10-03
NOISSUE - Invalidate thing key from cache on key update

bool3max:
When a Thing key is updated using the `.UpdateKey()` things service method, the previous key remains in the Redis cache and can still be used to authenticate the Thing.

This PR fixes that by invalidating the Thing keys from the cache after the key is updated.


duolok (approved):

majabirmancevic (approved):

manuio (approved):
LGTM

PR #868, opened 2025-10-03
MF-859 - Correct rollback usage in all services

### LINKED ISSUE #859: Use defer tx.Rollback()
Currently, when calling `BeginTxx`, the transaction is opened, and if a later return occurs before reaching `tx.Commit()` or `tx.Rollback()`, that transaction remains hanging and blocks the connection.

Put `defer tx.Rollback()` right after `BeginTxx`, so rollback will always be called unless Commit has already succeeded.

Do this in all services where necessary (like here: [things.Save](https://github.com/MainfluxLabs/mainflux/blob/dba2db22604b07276bc64256a083692b404db1d6/things/postgres/things.go#L36))

duolok:
Resolves #859


manuio (approved):
LGTM

bool3max (approved):

PR #877, opened 2025-10-09
MF-880 - Introduce PATCH Thing update endpoint and rework PUT update endpoint

### LINKED ISSUE #880: Introduce PATCH endpoint for updating Things

bool3max:
### Changes

1. Updates the `things.ThingRepository.Update()` method to support setting of all Thing columns
2. Removes the endpoints and service methods dedicated to updating a Thing's key (`PATCH /things/<thing-id>/key`, etc..)
3. Update the `PUT /things/<thing-id>` endpoint
    1. Remove the ability to update the Thing's Profile ID
    2. Add the ability to update the Thing's Key
    * In other words, the `PUT` endpoint is responsible for updating a Thing's name, metadata, and key
4. Introduces a `PATCH /things/<thing-id>` endpoint in the `things` service, that's able to update a specific Thing's belonging Profile or Group
   * When moving the Thing to another Group, it's mandatory to also supply a Profile ID as well, and the associated Profile must belong to the destination Group


things/tracing/things.go
```diff
@@ -13,10 +13,11 @@ import (
...
 const (
-	saveThing                = "save_thing"
-	saveThings               = "save_things"
-	updateThing              = "update_thing"
-	updateThingKey           = "update_thing_by_key"
+	saveThing   = "save_thing"
+	saveThings  = "save_things"
+	updateThing = "update_thing"
+	patchThing  = "patch_thing"
+
```
manuio: kill the blank line

api/openapi/things.yml
```diff
@@ -321,49 +321,47 @@ paths:
...
         - things
       parameters:
         - $ref: "#/components/parameters/ThingId"
+      requestBody:
+        $ref: "#/components/requestBodies/PatchThingReq"
       responses:
-        '204':
-          description: Thing removed.
+        '200':
+          description: Thing updated.
```
manuio, line 336: Should we use Thing patched?

pkg/sdk/go/things_test.go
```diff
@@ -30,6 +30,7 @@ const (
 	token       = email
 	wrongValue  = "wrong_value"
 	orgID       = "374106f7-030e-4881-8ab0-151195c29f92"
+	thingKey    = "tk1"
```
manuio: I'd use an UUID

things/api/http/things/requests.go
```diff
@@ -84,6 +84,34 @@ func (req updateThingReq) validate() error {
...
+func (req patchThingReq) validate() error {
+	if req.token == "" {
+		return apiutil.ErrBearerToken
+	}
+
+	if req.id == "" {
+		return apiutil.ErrMissingThingID
+	}
+
+	if req.ProfileID == "" && req.GroupID == "" {
```
manuio: Check only for ProfileID and return the corresponding error

things/api/http/things/requests.go
```diff
@@ -84,6 +84,34 @@ func (req updateThingReq) validate() error {
...
+		return apiutil.ErrMalformedEntity
+	}
+
+	// Making a request to change the Thing's Group requires specifying a Profile ID
+	if req.GroupID != "" && req.ProfileID == "" {
```
manuio: Same here for GroupID

things/api/logging.go
```diff
@@ -53,34 +53,34 @@ func (lm *loggingMiddleware) UpdateThing(ctx context.Context, token string, thin
 	return lm.svc.UpdateThing(ctx, token, thing)
 }
 
-func (lm *loggingMiddleware) UpdateThingsMetadata(ctx context.Context, token string, things ...things.Thing) (err error) {
+func (lm *loggingMiddleware) PatchThing(ctx context.Context, token string, thing things.Thing) (err error) {
```
manuio: It would have been better to modify next method instead of this one so the diffs are easier to understand

things/service.go
```diff
@@ -304,25 +302,42 @@ func (ts *thingsService) UpdateThing(ctx context.Context, token string, thing Th
...
 
-	prGrID, err := ts.getGroupIDByProfileID(ctx, thing.ProfileID)
-	if err != nil {
+	if err := ts.CanUserAccessThing(ctx, ar); err != nil {
 		return err
 	}
 
-	if prGrID != thGrID {
-		return errors.ErrAuthorization
+	if thing.ProfileID != "" {
```
manuio: Not sure that we need this check and in any case I don't think that this is correct. We have to check if user can access the thing and the given group plus check if profile is in the given group.

manuio (changes requested):

things/api/http/things/requests.go
bool3max: The check for the `GroupID` actually needs to go to the other PR - the one that adds support for changing Groups, so I'll remove the `GroupID` field from the `patchThingReq` struct in this PR.

things/api/http/things/requests.go
bool3max: ^ Same as above

api/openapi/things.yml
bool3max: I think "Thing updated" makes sense, in both cases (PUT/PATCH) you're updating the thing, just different aspects of it.

things/redis/streams.go
```diff
@@ -78,15 +78,30 @@ func (es eventStore) UpdateThing(ctx context.Context, token string, thing things
 	return nil
 }
 
-func (es eventStore) UpdateThingsMetadata(ctx context.Context, token string, things ...things.Thing) error {
```
manuio, line 81: Move `UpdateThingsMetadata` here

things/api/metrics.go
```diff
@@ -49,22 +49,22 @@ func (ms *metricsMiddleware) UpdateThing(ctx context.Context, token string, thin
 	return ms.svc.UpdateThing(ctx, token, thing)
 }
 
-func (ms *metricsMiddleware) UpdateThingsMetadata(ctx context.Context, token string, things ...things.Thing) error {
+func (ms *metricsMiddleware) UpdateThingGroupAndProfile(ctx context.Context, token string, thing things.Thing) error {
```
manuio: Move `UpdateThingsMetadata` here

manuio (changes requested):

things/api/logging.go
```diff
@@ -53,34 +53,34 @@ func (lm *loggingMiddleware) UpdateThing(ctx context.Context, token string, thin
 	return lm.svc.UpdateThing(ctx, token, thing)
 }
 
-func (lm *loggingMiddleware) UpdateThingsMetadata(ctx context.Context, token string, things ...things.Thing) (err error) {
+func (lm *loggingMiddleware) UpdateThingGroupAndProfile(ctx context.Context, token string, thing things.Thing) (err error) {
```
manuio: Define `UpdateThingsMetadata` here

things/api/metrics.go
manuio: Same here

manuio (changes requested):

manuio (approved):
LGTM

majabirmancevic (approved):

PR #882, opened 2025-10-13
MF-855 - JSON key filter

### LINKED ISSUE #855: Enhance readers to allow payload filtering by key

duolok:
Resolves #855

### DIFF
--- a/readers/api/transport.go
+++ b/readers/api/transport.go
@@ -35,6 +35,7 @@ const (
 	fromKey                = "from"
 	toKey                  = "to"
 	convertKey             = "convert"
+	jsonKeyKey             = "key"
 	aggIntervalKey         = "agg_interval"
 	aggTypeKey             = "agg_type"
 	aggFieldKey            = "agg_field"
@@ -123,6 +124,11 @@ func decodeListJSONMessages(_ context.Context, r *http.Request) (interface{}, er
 		return nil, err
 	}
 
+	key, err := apiutil.ReadStringQuery(r, jsonKeyKey, "")
+	if err != nil {
+		return nil, err
+	}
+
 	offset, err := apiutil.ReadUintQuery(r, apiutil.OffsetKey, apiutil.DefOffset)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


readers/api/transport.go
```diff
@@ -35,6 +35,7 @@ const (
 	fromKey                = "from"
 	toKey                  = "to"
 	convertKey             = "convert"
+	keyKey                 = "key"
```
majabirmancevic: I'd use `filterKey = "filter"`

readers/postgres/json.go
```diff
@@ -134,18 +135,44 @@ func (jr *jsonRepository) fmtCondition(rpm readers.JSONPageMetadata) string {
 		case "to":
 			condition = fmt.Sprintf(`%s %s created <= :to`, condition, op)
 			op = "AND"
+		case "key":
+			keyPath := buildPayloadFilterPath(rpm.Key)
+			condition = fmt.Sprintf(`%s %s %s IS NOT NULL`, condition, op, keyPath)
+			op = "AND"
+
```
manuio: kill the blank line

manuio (changes requested):

readers/api/transport.go
duolok: At first, I wanted to use `jsonKeyKey`, but `json` seemed a bit unnecessary. `filterKey` sounds ok, but the issue is that all of the other fields are filters too. `keyKey` didn't sound too good, but it's the most consistent with our current consts.

majabirmancevic: yeah, I get it..we use `name` to filter senml, but it's not intuitive in this case, so I think it's ok to keep "filter" for now

readers/messages.go
```diff
@@ -107,6 +107,7 @@ type JSONPageMetadata struct {
 	Protocol    string `json:"protocol,omitempty"`
 	From        int64  `json:"from,omitempty"`
 	To          int64  `json:"to,omitempty"`
+	Key         string `json:"key,omitempty"`
```
majabirmancevic: should we be consistent here, `Filter` instead of `Key`?

manuio (approved):
LGTM

majabirmancevic (approved):

PR #889, opened 2025-10-16
MF-886 - Assign/Unassign rules to things

### LINKED ISSUE #886: Enable assignment/unassignment rules for things
Add endpoints for assigning/removing rules to things, listing rules by thing, and listing thing IDs by rule

majabirmancevic:
Added endpoints:
- Assign rules: `POST /things/:id/rules` 
- Unassign rules: `PATCH /things/:id/rules` 
- List rules by thing: `GET /things/:id/rules` 
- List thing IDs by rule: `GET /rules/:id/things` 

Resolves #886

### DIFF
--- a/api/openapi/rules.yml
+++ b/api/openapi/rules.yml
@@ -48,6 +48,98 @@ paths:
           description: Database can't process request.
         '500':
           $ref: "#/components/responses/ServiceError"
+  /things/{thingId}/rules:
+    get:
+      summary: Retrieves rules assigned to a thing
+      description: Retrieves list of rules assigned to a specific thing identified by the provided ID.
+      tags:
+        - rules
+      parameters:
+        - $ref: "#/components/parameters/ThingId"
+      responses:
+        '200':
+          $ref: "#/components/responses/ListRulesRes"
+        '400':
+          description: Failed due to malformed query parameters.
+        '401':
+          description: Missing or invalid access token provided.
+        '403':
+          description: Failed to perform authorization over the entity.
+        '422':
+          description: Database can't process request.
+        '500':
+          $ref: "#/components/responses/ServiceError"
+    post:
+      summary: Assigns rules to a thing
+      description: Assigns rules to a specific thing identified by the provided ID.
+      tags:
+        - rules
+      parameters:
+        - $ref: "#/components/parameters/ThingId"
+      requestBody:
+        $ref: "#/components/requestBodies/ThingRulesReq"
+      responses:
+        '200':
+          description: Rules assigned successfully.
+        '400':
+          description: Failed due to malformed JSON.
+        '401':
+          description: Missing or invalid access token provided.
+        '403':
+          description: Failed to perform authorization over the entity.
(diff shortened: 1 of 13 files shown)


api/openapi/rules.yml
```diff
@@ -186,9 +183,6 @@ components:
         id:
           type: string
           example: "789e4567-e89b-12d3-a456-426614174abc"
-        profile_id:
-          type: string
-          example: "123e4567-e89b-12d3-a456-426614174000"
```
anna02272: Do we miss thing_ids in here?

api/openapi/rules.yml
```diff
@@ -318,8 +318,7 @@ components:
             items:
               $ref: "#/components/schemas/RuleResSchema"
           example:
-            - id: "84d1be5e-3511-4d15-aafa-b1bb4c20bca4"
-              profile_id: "123e4567-e89b-12d3-a456-426614174000"
```
anna02272: Do we miss thing_ids in here in example?

api/openapi/rules.yml
```diff
@@ -209,7 +203,7 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/Action"
-      required: [ id, profile_id, group_id, name, conditions, actions ]
+      required: [id, group_id, name, conditions, actions]
```
anna02272: And here?

rules/service.go
```diff
@@ -146,15 +150,21 @@ func (rs *rulesService) ViewRule(ctx context.Context, token, id string) (Rule, e
...
 
-	if _, err := rs.things.CanUserAccessProfile(ctx, &protomfx.UserAccessReq{Token: token, Id: rule.ProfileID, Action: things.Viewer}); err != nil {
+	if _, err := rs.things.CanUserAccessGroup(ctx, &protomfx.UserAccessReq{Token: token, Id: rule.GroupID, Action: things.Viewer}); err != nil {
 		return Rule{}, err
 	}
 
 	return rule, nil
 }
 
 func (rs *rulesService) UpdateRule(ctx context.Context, token string, rule Rule) error {
```
anna02272, line 171: Should we have option to update thing we created rule for?

api/openapi/rules.yml
majabirmancevic: In this implementation, thing_ids are not returned in responses, since there is a separate endpoint for retrieving rules per thing.
We might consider adding thing_ids to the response if that makes sense.

rules/service.go
majabirmancevic: yes, but I'd do it in a separate PR, because we should probably add new endpoints for Assign/Unassign, 
I wouldn't combine those updates

api/openapi/rules.yml
manuio: Yes please, add it to the response. Having only some of them is confusing.

anna02272 (approved):

manuio (approved):
LGTM

PR #891, opened 2025-10-17
Introduce support for HTML e-mails and multiple e-mail templates per Agent

### LINKED ISSUE #744: Enhance pkg/email with support for HTML templates and multiple templates per Agent
As we introduce support for e-mail notifications in more services and start sending end-user facing e-mails, it'd be great to support pretty HTML e-mails by facilitating [`html/template`](https://pkg.go.dev/html/template) in `pkg/email`.

We could also support multiple templates per one `email.Agent` - currently each Agent instance is tied to at most one raw-text template that always accepts only three "components": a Header, some Content, and a Footer.

bool3max:
This PR enhances the `pkg/email` package with support for sending HTML emails.

As pretty HTML emails can be quite complex (lots of inline CSS, for example), the package now also supports multiple "sub-templates" per Agent. All HTML e-mails are composed of one base template (`base.tmpl` by default) that's shared by all services and that includes common data (such as the overall HTML structure and lots of CSS style), within which a specific subtemplate is embedded once an e-mail is to be sent.

This allows services that require sending multiple different kinds of e-mails - for example the `users` service which sends password-reset request emails, e-mail verification emails, or platform invite e-mails - to define the structure of each one of those e-mails as a simple, clean, sub-template in its own file. For example, the template for the organization invite email in the `auth` service ([`docker/templates/auth/org_invite.tmpl`](https://github.com/bool3max/mainflux/blob/html-email-templates/docker/templates/auth/org_invite.tmpl)) could look like so:

```
{{ define "content" }}
    <p>You've been invited to join an Organization:</p>
    <ul>
        <li>Organization: {{ .OrgName }}</li>
        <li>Role: {{ .Role }}</li>
    </ul>
    
    <a href="{{ .InviteLink }}">Click here to view the invite</a>
{{ end }}
```

image

## How it works

* The base template (`base.tmpl`) contains common data shared by all templates (header, footer, CSS styles, etc...)
* It references a `{{ template "content" }}`
* Each sub-template defines this `"content"` block within itself
* When sending an e-mail, the call to [`email.Agent.Send()`](https://github.com/bool3max/mainflux/blob/62bde7295499a4aeb526eec9030120663fc9d803/pkg/email/email.go#L80) accepts a `TemplateName` that corresponds to the filename of the desired sub-template.
    * It parses the sub-template and injects it into the base template
    * It then renders the final template, passing in any user-defined variables that may be required by the sub-template

## Compromises

There's one issue: 

It would be really nice if golang's `html/template` package allowed us to parse a bunch of template files upfront, and then render a specific one at will. Something along the lines of the following inside a "base" template:

```
<html>
<head>
    <style> /* bunch of inline CSS... */ </style>
</head>
	{{ template .TemplateName . }}
	            ^
	            |
		template to render
		passed as variable during base template execution
</html>
```

Unfortunately, that doesn't work. The first argument to the `{{ template }}` action must be a static string, and it must reference an already-parsed template. Knowing this, we have two options that allow us to have different e-mail templates in different files, while having one "base" template.

**Option 1**. involves having each service (auth, users, smtp-notifier, etc..) define its own "base" template file, that would within itself include an if-else chain that would pick the sub-template to include based on an external variable:

```
{{ if eq .Action "platform_invite"}}
    {{ template "platform_invite_template_name" . }}
{{ else if eq .Action "password_reset" }}
    {{ template "password_reset_template_name" . }}
{{ end }}
```

... where each of the sub-templates comes from its own file that would `{{ define "own_template_name" }}`. This would allow us to parse all needed templates *once*, when `Agent` is first created (i.e. when the service first starts up), and then execute/render them at will. However, the big glaring issue with this approach is that it's extremely unmaintainable, as each service needs a copy of the base template. Whenever we'd want to make a change to the base template, we'd have to edit the base template specific to each service. Very unmaintainable.

**Option 2**. is the one I went with, and it's to have one master base template that references a `{{ template "content" }}`, which is then extended by a specific sub-template (that includes a `{{ define "content" }}`) parsed at 'runtime', whenever an e-mail is being sent. This does introduce a performance cost, but I think it's very negligible as sub-templates themselves are very little in size. This approach is a lot cleaner and easier to maintain as there only exists one "base" template shared amongst all services, and it doesn't require you to list all possible sub-templates within it in advance.

Resolves #744.


anna02272:
I'm thinking about maybe making this email a bit more professional and less simple.

Maybe we could use a **button instead of a plain link**?

We could also add a **picture of the Mainflux logo** to make it feel more polished, something like what we have on the Mainflux website:
Screenshot from 2025-10-17 14-15-45

Also, we could include a **footer in the template** like:

```
Best regards,
The Mainflux Team
Email: infos@mainflux.com
Website: mainflux.com
```

Is this possible with the current templates?
What do you think?

bool3max:
Oh of course, the ones I included in this PR are just placeholders, pretty much to show support for HTML e-mails. Later on we can actually make them pretty!

docker/templates/auth/org_invite.tmpl
```diff
@@ -0,0 +1,9 @@
+{{ define "content" }}
+    <p>You've been invited to join an Organization:</p>
+    <ul>
+        <li>Organization: {{ .OrgName }}</li>
+        <li>Role: {{ .Role }}</li>
+    </ul>
+    
+    <a href="{{ .InviteLink }}">Click here to view the invite</a>
+{{ end }}
```
anna02272: No newline at end of file

docker/templates/base.tmpl
```diff
@@ -0,0 +1,12 @@
...
+<html lang="en">
+<head>
+    <title>{{ .Title }}</title>
+</head>
+<body>
+    <h1>MainfluxLabs</h1>
+    {{ template "content" . }}
+</body>
+</html>
+{{ end }}
```
anna02272: No newline at end of file

docker/templates/smtp/notification.tmpl
```diff
@@ -0,0 +1,4 @@
+{{ define "content" }}
+    <p>A publisher with an id {{ .PublisherID }} sent the message over {{ .Protocol }} with the following values:</p>
+    <p>{{ .MessageContent }}</p>
+{{ end }}
```
anna02272: No newline at end of file

docker/templates/users/email_verification.tmpl
```diff
@@ -0,0 +1,5 @@
+{{ define "content" }}
+    <p>Navigate to the following URL to verify your e-mail address and complete MainfluxLabs account registration:</p>
+
+    <a href="{{ .RedirectURL }}">Click here to verify your e-mail address.</a>
+{{ end }}
```
anna02272: No newline at end of file

docker/templates/users/password_reset.tmpl
```diff
@@ -0,0 +1,5 @@
+{{ define "content" }}
+    <p>You've requested a password reset of your MainfluxLabs user account.</p>        
+
+    <a href="{{ .RedirectURL }}">Click here to reset your password.</a>
+{{ end }}
```
anna02272: No newline at end of file

docker/templates/users/platform_invite.tmpl
```diff
@@ -0,0 +1,5 @@
+{{ define "content" }}
+    <h3>You've been invite to join MainfluxLabs!</h3>
+
+    <a href="{{ .RedirectURL }}">Click here to create an account.</a>
+{{ end }}
```
anna02272: No newline at end of file

docker/templates/users/platform_invite.tmpl
```diff
@@ -0,0 +1,5 @@
+{{ define "content" }}
+    <h3>You've been invite to join MainfluxLabs!</h3>
```
anna02272: invited

anna02272 (approved):

manuio (approved):
LGTM

PR #896, opened 2025-10-20
MF-887 - Correctly translate failed gRPC service calls to HTTP responses

### LINKED ISSUE #887: Fix HTTP 500 returned for gRPC errors
The `errors.Contains` helper function did not properly recognize gRPC `PermissionDenied` (e.g., `rpc error: code = PermissionDenied desc = failed to perform authorization over the entity.`) and other wrapped errors, causing other services to return `500 Internal Server Error ` instead of `403 Forbidden`.

bool3max:
Resolves #887 by testing the error that arrives in the `encodeError` function of each service's HTTP layer for a gRPC `status.Status` error, mapping its `Code` to an appropriate HTTP response status code, and constructing the response JSON error using its attached `.Message()`.

### DIFF
--- a/pkg/apiutil/transport.go
+++ b/pkg/apiutil/transport.go
@@ -125,17 +125,14 @@ func EncodeGRPCError(st *status.Status, w http.ResponseWriter) {
 	case codes.OK:
 		w.WriteHeader(http.StatusOK)
 
-	case codes.Canceled:
-		w.WriteHeader(http.StatusRequestTimeout)
-
-	case codes.Unknown:
+	case codes.Internal:
 		w.WriteHeader(http.StatusInternalServerError)
 
 	case codes.InvalidArgument:
 		w.WriteHeader(http.StatusBadRequest)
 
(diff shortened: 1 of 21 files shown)


pkg/apiutil/transport.go
```diff
@@ -185,14 +230,29 @@ func EncodeError(err error, w http.ResponseWriter) {
...
 		w.WriteHeader(http.StatusInternalServerError)
 	}
 }
 
 func WriteErrorResponse(err error, w http.ResponseWriter) {
+	var errorMessage string
+
 	if errorVal, ok := err.(errors.Error); ok {
+		errorMessage = errorVal.Msg()
+	} else if st, ok := status.FromError(err); ok {
```
manuio: Use a switch case instead

manuio (changes requested):

manuio (approved):
LGTM

majabirmancevic (approved):

PR #900, opened 2025-10-23
MF-897 - Add group check to CanUserAccessGroup

### LINKED ISSUE #897: Missing check if group exists in CanUserAccessGroup
The root admin can create some entities (notifiers, rules,..) for a group that does not exist, because there is no check if the group exists.
Add a required check in the CanUserAccessGroup gRPC method.

majabirmancevic:
Resolves #897

`isAdmin` check moved from the top of the function to avoid unnecessary gRPC calls for regular users.

### DIFF
--- a/things/groups.go
+++ b/things/groups.go
@@ -299,10 +299,6 @@ func (ts *thingsService) ViewGroupByThing(ctx context.Context, token string, thi
 }
 
 func (ts *thingsService) canAccessGroup(ctx context.Context, token, groupID, action string) error {
-	if err := ts.isAdmin(ctx, token); err == nil {
-		return nil
-	}
-
 	user, err := ts.auth.Identify(ctx, &protomfx.Token{Value: token})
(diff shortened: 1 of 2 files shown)


things/groups.go
```diff
@@ -317,34 +313,32 @@ func (ts *thingsService) canAccessGroup(ctx context.Context, token, groupID, act
...
+				return nil
+			}
 			return err
 		}
 		role = r
-
-		if err := ts.groupCache.SaveGroupMembership(ctx, gm.GroupID, gm.MemberID, r); err != nil {
-			return err
-		}
+		_ = ts.groupCache.SaveGroupMembership(ctx, gm.GroupID, gm.MemberID, r)
```
manuio, line 323: We can consider removing cache errors or using a logger

manuio (approved):
LGTM

anna02272 (approved):

PR #934, opened 2025-11-18
MF-933 - Use configurable consumer groups for Redis subscriber

### LINKED ISSUE #933: Make Redis subscriber use configurable consumer groups
The current Redis Streams subscriber implementation uses a single hard-coded consumer group (`mainflux`) for all services.
This causes subscribers to compete for messages, meaning only one service receives an event instead of all services getting it.

Extend the Redis subscriber to accept a consumer group name as a constructor argument, allowing each service to define its own group and therefore receive all events independently.

majabirmancevic:
Resolves #933


manuio (approved):
LGTM

PR #937, opened 2025-11-19
MF-936 - Add configurable date format, message order, and payload columns to reports

### LINKED ISSUE #936: Improve reports with configurable date format, message order, and payload fields
Date format should be configurable
Order of messages should be configurable ASC/DESC
We should be able to select the hours in a day for the from/to filters
Payload could be split in columns per field

anna02272:
Resolves #936


readers/api/converters.go
```diff
@@ -149,19 +172,99 @@ func getValue(ptr interface{}, defaultValue string) string {
...
-func GenerateJSON(page readers.MessagesPage) ([]byte, error) {
-	if page.Total == 0 {
+func formatTimeNs(ns int64, timeFormat string) string {
+	if strings.ToLower(timeFormat) == "rfc3339" {
+		return time.Unix(0, ns).UTC().Format(time.RFC3339)
+	}
+	return fmt.Sprintf("%v", ns)
+}
+
+func Flatten(m map[string]interface{}, prefix string) map[string]interface{} {
```
bool3max: We can use `any` in place of `interface{}` throughout the entire function

bool3max (changes requested):

readers/api/converters.go
```diff
@@ -149,19 +172,99 @@ func getValue(ptr interface{}, defaultValue string) string {
 	return defaultValue
 }
 
-func GenerateJSON(page readers.MessagesPage) ([]byte, error) {
-	if page.Total == 0 {
+func formatTimeNs(ns int64, timeFormat string) string {
+	if strings.ToLower(timeFormat) == "rfc3339" {
```
majabirmancevic, line 176: Why are we only checking `rfc3339`?

readers/api/converters.go
```diff
@@ -149,19 +172,99 @@ func getValue(ptr interface{}, defaultValue string) string {
 	return defaultValue
 }
 
-func GenerateJSON(page readers.MessagesPage) ([]byte, error) {
-	if page.Total == 0 {
+func formatTimeNs(ns int64, timeFormat string) string {
```
majabirmancevic: formatTime

readers/api/converters.go
```diff
@@ -149,19 +172,99 @@ func getValue(ptr interface{}, defaultValue string) string {
...
+				result[nk] = nv
+			}
+		default:
+			result[key] = v
+		}
+	}
+	return result
+}
+
+func GenerateJSONFromJSON(page readers.JSONMessagesPage) ([]byte, error) {
```
majabirmancevic: let's find a better name

readers/api/requests.go
```diff
@@ -107,14 +122,18 @@ func (req backupJSONMessagesReq) validate() error {
 		return err
 	}
 
+	if req.pageMeta.Dir != "" &&
+		req.pageMeta.Dir != apiutil.AscDir && req.pageMeta.Dir != apiutil.DescDir {
```
majabirmancevic: Can a helper be used for this check?

readers/api/converters.go
```diff
@@ -75,53 +76,75 @@ func GenerateCSVFromSenML(page readers.MessagesPage) ([]byte, error) {
...
 
-	writer.Flush()
-	if err := writer.Error(); err != nil {
+	header := append(jsonHeader, payload...)
+	if err := writer.Write(header); err != nil {
 		return nil, err
 	}
 
-	return buf.Bytes(), nil
+	for i, raw := range page.MessagesPage.Messages {
```
majabirmancevic, line 116: a bit redundant `page.MessagesPage.Messages`

readers/api/converters.go
anna02272: We just need something human readable, and Manuel has decided that support for rfc3339 is enough

majabirmancevic: ok, so what would a _configurable date format_ mean? @manuio

readers/api/converters.go
```diff
@@ -149,19 +172,99 @@ func getValue(ptr interface{}, defaultValue string) string {
...
 
 	return data, nil
 }
 
+func GenerateJSONFromSenML(page readers.SenMLMessagesPage) ([]byte, error) {
+	if page.MessagesPage.Total == 0 {
+		return []byte("[]"), nil
+	}
+
+	out := make([]map[string]interface{}, 0, len(page.MessagesPage.Messages))
```
bool3max: All of these `interface{}` can be just `any`, they're totally equivalent but more readable. (see also #916)

bool3max (changes requested):

readers/api/converters.go
anna02272: I actually saw that we get message `interface{} can be replaced by any`. But also we use it like this in literally whole project

bool3max: Yup though I opened a PR to replace all occurrences by `any`: https://github.com/MainfluxLabs/mainflux/pull/947. So going forward in future PRs (and this one) we can just use any by default.

anna02272: Okay okay.  I didn't see link to the issue.

bool3max:
`GenerateCSVFromJSON` previously just dumped the entire JSON payload as a JSON-string in the last CSV column (`payload`). Now it attempts to flatten the message objects, which works great and is very readable for SenML or other simple nested objects but seems to break on arrays, e.g. the following message payload:

```json
{
    "v": 3,
    "data": [
        {
            "name": "abc def",
            "num": 33
        },
        {
            "name": "ghi ccc",
            "num": 44
        }
    ]
}
```

... results in the following CSV row in the report:

```csv
created,subtopic,publisher,protocol,v,data
1764016522583648513,,6eb8f7d9-b724-4bfe-b7f0-78643d2e3646,http,3,[map[name:abc def num:33] map[name:ghi ccc num:44]]
```

It ends up spitting out Golang-formatted maps and slices which aren't readable or easily parsable. Maybe it's better to keep the old "JSON-string-in-CSV" approach for JSON messages (even though it sort of defeats the purpose of generating a CSV report?) and use this new one for SenML only?

anna02272:
> `GenerateCSVFromJSON` previously just dumped the entire JSON payload as a JSON-string in the last CSV column (`payload`). Now it attempts to flatten the message objects, which works great and is very readable for SenML or other simple nested objects but seems to break on arrays, e.g. the following message payload:
> 
> ```json
> {
>     "v": 3,
>     "data": [
>         {
>             "name": "abc def",
>             "num": 33
>         },
>         {
>             "name": "ghi ccc",
>             "num": 44
>         }
>     ]
> }
> ```
> 
> ... results in the following CSV row in the report:
> 
> ```
> created,subtopic,publisher,protocol,v,data
> 1764016522583648513,,6eb8f7d9-b724-4bfe-b7f0-78643d2e3646,http,3,[map[name:abc def num:33] map[name:ghi ccc num:44]]
> ```
> 
> It ends up spitting out Golang-formatted maps and slices which aren't readable or easily parsable. Maybe it's better to keep the old "JSON-string-in-CSV" approach for JSON messages (even though it sort of defeats the purpose of generating a CSV report?) and use this new one for SenML only?


I wouldn’t go back to the old approach, since the requirement of issue was to split the payload into columns per field. 

I initially tried a version where each element in a list became its own set of columns, but that quickly started to look line an overkill. For long lists, we’d need list[0].field1, list[0].field2, list[1].field1… and so on, which would make the table very long and hard to read.

I think it’s fine to keep lists in a single column, we stored whole payload like that earlier. 

It’s the Flatten method Manuel sent me, which has been used somewhere before.

For SenML messages, this isn’t an issue anyway, since they only contain key value pairs, we don't have payload.

bool3max:
Right but I think that the main issue is the use of Golang-formatted values for any moderately-complex fields. For example if the message payload somewhere within contains an array of strings such as: `["abc de", "fg h i"`], the column in the final CSV will contain:

```
[abc de fg h i]
```

Which makes no sense. Is it a slice of one string with 4 spaces (`["abc de fg h i"]`), or maybe a slice of 3 strings (`["abc", "de fg", "h i"]`), or any combination of those? You can't tell.

manuio (approved):
LGTM

majabirmancevic (approved):

bool3max (approved):

PR #943, opened 2025-11-21
MF-235 - Implement OAuth 2.0 SSO with Google and GitHub Providers

### LINKED ISSUE #235: Implement OAuth 2.0 SSO Login Support
Support OAuth 2.0-based SSO by integrating third-party providers (e.g., Google, GitHub) for user authentication. Handle token exchange, user profile retrieval, and session management.

anna02272:
Resolves #235 

**Setup for Google and GitHub SSO**

To make SSO work, we need to configure Google and GitHub developer clients.

**Google:**

1. Go to [Google Cloud Console](https://console.cloud.google.com/), create a new project.
2. Navigate to **APIs & Services → Credentials → Create OAuth Client**.
3. Set **Authorized redirect URI** to `http://localhost/users/oauth/google/callback` (this should match `MF_GOOGLE_REDIRECT_URL`).
4. After creation, copy the **Client ID** and **Client Secret** into `MF_GOOGLE_CLIENT_ID` and `MF_GOOGLE_CLIENT_SECRET`.

**GitHub:**

1. Go to **Settings → Developer Settings → OAuth Apps → New OAuth App**.
2. Set the **Authorization callback URL** to `http://localhost/users/oauth/github/callback` (`MF_GITHUB_REDIRECT_URL`).
3. Copy the **Client ID** and **Client Secret** into `MF_GITHUB_CLIENT_ID` and `MF_GITHUB_CLIENT_SECRET`.

**Example `.env`:**

```
MF_GOOGLE_CLIENT_ID=google-client-id.apps.googleusercontent.com
MF_GOOGLE_CLIENT_SECRET=google-client-secret
MF_GOOGLE_REDIRECT_URL=/users/oauth/google/callback

MF_GITHUB_CLIENT_ID=github-client-id
MF_GITHUB_CLIENT_SECRET=github-client-secret
MF_GITHUB_REDIRECT_URL=/users/oauth/github/callback
```


users/api/http/requests.go
```diff
@@ -66,6 +68,41 @@ func (req registerByInviteReq) validate() error {
 	return nil
 }
 
+type oauthLoginReq struct {
+	provider string
+}
+
+func (req oauthLoginReq) validate() error {
+	if req.provider == "" || (req.provider != GoogleProvider && req.provider != GitHubProvider) {
```
bool3max: Don't need the `req.provider == ""` check

users/api/http/requests.go
```diff
@@ -66,6 +68,41 @@ func (req registerByInviteReq) validate() error {
...
+type oauthCallbackReq struct {
+	provider      string
+	code          string
+	state         string
+	originalState string
+	verifier      string
+}
+
+func (req oauthCallbackReq) validate() error {
+	if req.provider == "" || (req.provider != GoogleProvider && req.provider != GitHubProvider) {
```
bool3max: Same here no need for `req.provider == ""`

bool3max (changes requested):

users/service.go
```diff
@@ -422,6 +447,106 @@ func (svc usersService) Login(ctx context.Context, user User) (string, error) {
...
+				break
+			}
+		}
+
+		if email == "" && len(emails) > 0 {
+			email = emails[0].Email
+		}
+	}
+
+	user, err := svc.users.RetrieveByEmail(ctx, email)
```
bool3max: You can "un-nest" the error handling here and avoid the if-else:

```
user, err := svc.users.RetrieveByEmail(ctx, email)
if err != nil {
    if !errors.Contains(err, dbutil.ErrNotFound) {
        return "", err
    }

    // save user etc..
   ...
}

...
```

bool3max (changes requested):

users/service.go
```diff
@@ -422,6 +447,106 @@ func (svc usersService) Login(ctx context.Context, user User) (string, error) {
...
+	if err != nil {
+		if errors.Contains(err, dbutil.ErrNotFound) {
+			uid, _ := svc.idProvider.ID()
+			user = User{
+				ID:     uid,
+				Email:  email,
+				Status: EnabledStatusKey,
+			}
+			if _, err := svc.users.Save(ctx, user); err != nil {
+				return "", err
```
bool3max: The `User` instance has no set `Password` field, so it defaults to an empty string `""`. You've dropped the `NULL` constraint on the `password` field in the users database ([here](https://github.com/MainfluxLabs/mainflux/pull/943/files#diff-b5ab9435df40fd1cb78deca246b580e88029439e5fde2b8f593d716ea9ca8e6fR100)), but saving this User here doesn't actually insert a row with `password=NULL`, but instead with `password=''` (empty string) I'm pretty sure.

To actually insert `NULL` on an empty password you can use the `sql.NullString` type for the `Password` field in [`dbUser`](https://github.com/MainfluxLabs/mainflux/blob/aa0225fc0cf62be05a548891505e914f743e3227/users/postgres/users.go#L279).

I did something very similar for Org invites - you can see [here](https://github.com/MainfluxLabs/mainflux/blob/aa0225fc0cf62be05a548891505e914f743e3227/auth/postgres/invites.go#L465).

bool3max (changes requested):

users/service.go
```diff
@@ -422,6 +447,108 @@ func (svc usersService) Login(ctx context.Context, user User) (string, error) {
...
+		oauthCfg = svc.githubOAuth
+	}
+
+	verifier = oauth2.GenerateVerifier()
+	state = generateRandomState()
+	redirectURL = oauthCfg.AuthCodeURL(state, oauth2.AccessTypeOffline, oauth2.S256ChallengeOption(verifier))
+	return state, verifier, redirectURL
+}
+
+func (svc usersService) OAuthCallback(ctx context.Context, provider, code, verifier string) (string, error) {
```
bool3max: Reminder that we should authenticate based on the unique user ID returned by the provider rather than the email.

users/api/http/users/endpoint.go
```diff
@@ -216,7 +216,10 @@ func oauthLoginEndpoint(svc users.Service) endpoint.Endpoint {
 			return nil, err
 		}
 
-		state, verifier, redirectURL := svc.OAuthLogin(req.provider)
+		state, verifier, redirectURL, err := svc.OAuthLogin(req.provider)
```
manuio: Use a structure for `state, verifier, redirectURL`

manuio (changes requested):

users/service_test.go
```diff
@@ -706,48 +705,49 @@ func TestOAuthCallback(t *testing.T) {
 
 	verifier := oauth2.GenerateVerifier()
 
-	cases := map[string]struct {
+	cases := []struct {
+		desc                string
 		provider            string
 		code                string
 		verifier            string
 		wantErr             error
```
manuio: Is this naming consistent with other tests?

manuio (changes requested):

cmd/users/main.go
```diff
@@ -329,7 +381,7 @@ func newService(db *sqlx.DB, tracer opentracing.Tracer, ac protomfx.AuthServiceC
 
 	idProvider := uuid.New()
 
-	svc := users.New(userRepo, verificationRepo, platformInvitesRepo, c.inviteDuration, c.emailVerifyEnabled, c.selfRegisterEnabled, hasher, ac, svcEmailer, idProvider)
+	svc := users.New(userRepo, verificationRepo, platformInvitesRepo, identityRepo, c.inviteDuration, c.emailVerifyEnabled, c.selfRegisterEnabled, hasher, ac, svcEmailer, idProvider, c.googleOauthConfig, c.githubOauthConfig, c.urls)
```
majabirmancevic: we could pass `c` to reduce the number of args

pkg/apiutil/errors.go
```diff
@@ -179,4 +179,13 @@ var (
 
 	// ErrInvalidThingKeyType indicates an invalid or missing type of thing authentication key
 	ErrInvalidThingKeyType = errors.New("invalid thing key type")
+
+	// ErrMissingProvider indicates an invalid or missing provider
+	ErrMissingProvider = errors.New("missing or invalid provider")
```
majabirmancevic: `ErrInvalidProvider` (to be consistent with others)

pkg/apiutil/errors.go
```diff
@@ -179,4 +179,13 @@ var (
...
+
+	// ErrMissingProviderCode indicates a missing provider code
+	ErrMissingProviderCode = errors.New("missing provider code")
+
+	// ErrMissingState indicates an invalid or missing state
+	ErrMissingState = errors.New("missing or invalid state")
```
majabirmancevic: `ErrInvalidState`

users/api/http/backup/endpoint.go
```diff
@@ -30,9 +30,9 @@ func restoreEndpoint(svc users.Service) endpoint.Endpoint {
 			return nil, err
 		}
 
-		admin, users := buildBackup(req)
+		admin, users, identities := buildBackup(req)
 
-		err := svc.Restore(ctx, req.token, admin, users)
+		err := svc.Restore(ctx, req.token, admin, users, identities)
 		if err != nil {
```
majabirmancevic: inline err

users/postgres/identity.go
```diff
@@ -0,0 +1,104 @@
...
+	return &identityRepository{
+		db: db,
+	}
+}
+func (ir identityRepository) Save(ctx context.Context, identity users.Identity) error {
+	q := `INSERT INTO oauth_identities (user_id, provider, provider_user_id) VALUES (:user_id, :provider, :provider_user_id)`
+
+	dbi := toDBIdentity(identity)
+
+	_, err := ir.db.NamedExecContext(ctx, q, dbi)
```
majabirmancevic: inline err

users/service.go
```diff
@@ -416,12 +446,210 @@ func (svc usersService) Login(ctx context.Context, user User) (string, error) {
...
 
+func (svc usersService) OAuthLogin(provider string) (data OAuthLoginData, err error) {
+	var oauthCfg oauth2.Config
+	switch provider {
+	case GoogleProvider:
+		oauthCfg = svc.googleOAuth
+	case GitHubProvider:
+		oauthCfg = svc.githubOAuth
+	default:
+		return OAuthLoginData{}, errors.ErrAuthorization
```
majabirmancevic: ErrAuthentication?

users/service.go
```diff
@@ -416,12 +446,210 @@ func (svc usersService) Login(ctx context.Context, user User) (string, error) {
...
+	var email, providerUserID string
+	var err error
+
+	switch data.Provider {
+	case GoogleProvider:
+		email, providerUserID, err = svc.fetchGoogleUser(ctx, data.Code, data.Verifier)
+	case GitHubProvider:
+		email, providerUserID, err = svc.fetchGitHubUser(ctx, data.Code, data.Verifier)
+	default:
+		return "", errors.ErrAuthorization
```
majabirmancevic: same here

users/service_test.go
```diff
@@ -199,7 +223,7 @@ func TestViewUser(t *testing.T) {
 		err    error
 	}{
 		"view user with authorized token": {
-			user:   user,
+			user:   users.User{ID: user.ID, Email: user.Email, Metadata: user.Metadata, Status: user.Status, Role: user.Role},
```
majabirmancevic: why not `user`?

users/service_test.go
```diff
@@ -240,12 +264,12 @@ func TestViewProfile(t *testing.T) {
...
 	}{
 		"valid token's user info": {
-			user:  user,
+			user:  users.User{ID: user.ID, Email: user.Email, Metadata: user.Metadata, Status: user.Status, Role: user.Role},
 			token: token,
 			err:   nil,
 		},
 		"valid token's admin info": {
-			user:  admin,
+			user:  users.User{ID: admin.ID, Email: admin.Email, Metadata: admin.Metadata, Status: admin.Status, Role: admin.Role},
```
majabirmancevic: and `admin` here?

users/service_test.go
```diff
@@ -630,3 +654,104 @@ func TestValidatePlatformInvite(t *testing.T) {
...
+
+func TestOAuthLogin(t *testing.T) {
+	svc := newService()
+
+	cases := map[string]struct {
+		provider      string
+		stateNotEmpty bool
+		err           error
+	}{
+		"oauth login with google provider":  {users.GoogleProvider, true, nil},
```
majabirmancevic: let's use the same style as in TestOAuthCallback
```
{
	desc:      "..",
	provider:  "..",
```

users/tracing/users.go
```diff
@@ -48,11 +49,19 @@ func (urm userRepositoryMiddleware) Save(ctx context.Context, user users.User) (
...
-func (urm userRepositoryMiddleware) UpdateUser(ctx context.Context, user users.User) error {
+func (urm userRepositoryMiddleware) Update(ctx context.Context, user users.User) error {
 	span := dbutil.CreateSpan(ctx, urm.tracer, updateUser)
 	defer span.Finish()
 	ctx = opentracing.ContextWithSpan(ctx, span)
 
+	return urm.repo.Update(ctx, user)
+}
+
+func (urm userRepositoryMiddleware) UpdateUser(ctx context.Context, user users.User) error {
```
majabirmancevic, line 60: `UpdateUserMetadata`?
#later

users/service_test.go
anna02272: Because user / admin contain more fields than the method under test returns, especially Password. But if I removed Password from the user, admin, etc., all tests that first Login to get a token would start failing.

manuio: Can we create `oauth2User` maybe?

users/tracing/users.go
manuio: Create an issue

users/service_test.go
anna02272: > Can we create `oauth2User` maybe?

That behavior is the same whether the user originally logged in via Email/password or OAuth.

manuio: I meant, why can't we create a variable `oauth2User = users.User{ID: user.ID, Email: user.Email, Metadata: user.Metadata, Status: user.Status, Role: user.Role}`?

users/service_test.go
```diff
@@ -36,15 +36,16 @@ const (
...
-	unauthUser      = users.User{Email: "unauth-user@example.com", ID: "6a32810a-4451-4ae8-bf7f-4b1752856eef", Password: "password"}
-	selfRegister    = users.User{Email: "self-register@example.com", Password: "password"}
-	registerUser    = users.User{Email: "register-user@example.com", ID: "574106f7-030e-4881-8ab0-151195c29f95", Password: "password"}
-	user            = users.User{Email: "user@example.com", ID: "574106f7-030e-4881-8ab0-151195c29f96", Password: "password"}
-	noPasswordUser  = users.User{Email: "user@example.com", ID: "574106f7-030e-4881-8ab0-151195c29f96"}
-	nonExistingUser = users.User{Email: "non-ex-user@example.com", Password: "password"}
-	usersList       = []users.User{admin, registerUser, user, unauthUser}
-	host            = "example.com"
+	admin            = users.User{Email: "admin@example.com", ID: "574106f7-030e-4881-8ab0-151195c29f94", Role: "root", Password: "password"}
+	noaPasswordAdmin = users.User{Email: "admin@example.com", ID: "574106f7-030e-4881-8ab0-151195c29f94", Role: "root"}
```
manuio: `noaPasswordAdmin`

manuio (changes requested):

manuio (approved):
LGTM

majabirmancevic (approved):

bool3max (approved):

PR #946, opened 2025-11-25
MF-913 - Remove references to types from deprecated github.com/golang/protobuf module

### LINKED ISSUE #913: Remove references to deprecated github.com/golang/protobuf module
Multiple of our own gRPC client/server packages reference both the _deprecated_ `github.com/golang/protobuf` module and its types (i.e. `empty.Empty`), as as well as the newer `google.golang.org/protobuf` module and its types (`emptypb.Empty`).

bool3max:
The references were mostly to the `empty.Empty` type in the gRPC client/server code. 

However, we still cannot remove the `github.com/golang/protobuf` module as a dependency because **the protobuf compiler we use ([`protoc-gen-gofast`](https://github.com/gogo/protobuf)) emits code that depends on it**. Though as the compiler itself is unmaintained since ~2021., we can think about moving to [`protoc-gen-go`](https://pkg.go.dev/google.golang.org/protobuf@v1.36.10/cmd/protoc-gen-go)?

### DIFF
--- a/auth/api/grpc/client.go
+++ b/auth/api/grpc/client.go
@@ -11,9 +11,9 @@ import (
 	"github.com/go-kit/kit/endpoint"
 	kitot "github.com/go-kit/kit/tracing/opentracing"
 	kitgrpc "github.com/go-kit/kit/transport/grpc"
-	"github.com/golang/protobuf/ptypes/empty"
 	opentracing "github.com/opentracing/opentracing-go"
 	"google.golang.org/grpc"
+	"google.golang.org/protobuf/types/known/emptypb"
 )
 
 const (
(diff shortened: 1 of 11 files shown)


duolok (approved):
LGTM

manuio:
> The references were mostly to the `empty.Empty` type in the gRPC client/server code.
> 
> However, we still cannot remove the `github.com/golang/protobuf` module as a dependency because **the protobuf compiler we use ([`protoc-gen-gofast`](https://github.com/gogo/protobuf)) emits code that depends on it**. Though as the compiler itself is unmaintained since ~2021., we can think about moving to [`protoc-gen-go`](https://pkg.go.dev/google.golang.org/protobuf@v1.36.10/cmd/protoc-gen-go)?

Good catch, let's use protoc-gen-go

manuio (approved):
LGTM

Issue #954, opened 2025-11-28
Consume things events in all services where necessary


majabirmancevic:
When the **Things** service emits events (listed below), other services must perform appropriate actions on dependent entities.

- ### `thing.remove`

### Rules service
- [x] Remove rules assigned to the thing

### Alarms service
- [x] Remove alarms belonging to the thing

### MQTT-adapter service
- [x] Remove subscriptions by thing

---

- ### `group.remove`

>  **Group Things / Profiles / Memberships are deleted automatically by PostgreSQL FK constraints.**

### Notifiers service
- [x] Remove notifiers belonging to the group

### Alarms service
- [x] Remove alarms for things in the group

### Rules service
- [x] Remove rules belonging to the group

### MQTT-adapter service
- [x] Remove subscriptions by group

majabirmancevic:
@bool3max Consider consuming events for updating the Thing Group.

majabirmancevic:
> [@bool3max](https://github.com/bool3max) Consider consuming events for updating the Thing Group.

@bool3max Please check this and open a sub-issue if necessary.

PR #955, opened 2025-11-28
MF-954 - Move event consts to pkg

majabirmancevic:
To avoid creating the same constants in every service that consumes these events, they are moved to `pkg` and can be used from there.


majabirmancevic:
#954  **_Reminder_**:
Find a more consistent value for this const later
`ThingUpdateGroupAndProfile = thingPrefix + "update_group_and_profile"`

duolok (dismissed):

pkg/events/utils.go
```diff
@@ -0,0 +1,27 @@
...
+	ThingRemove                = thingPrefix + "remove"
+
+	ProfileCreate = profilePrefix + "create"
+	ProfileUpdate = profilePrefix + "update"
+	ProfileRemove = profilePrefix + "remove"
+
+	GroupRemove = groupPrefix + "remove"
+)
+
+func Read(event map[string]interface{}, key, def string) string {
```
bool3max: Can use `any` instead of `interface{}` here and pretty much everywhere else in the PR

bool3max (changes requested):

manuio (dismissed):
LGTM

duolok (approved):
LGTM

anna02272 (approved):

manuio (approved):
LGTM

bool3max (approved):

PR #960, opened 2025-12-01
MF-959 - Refactor Redis event store middleware to use interface embedding

### LINKED ISSUE #959: Refactor Redis event store middleware to use interface embedding
The Redis event store middleware currently implements all methods from the `things.Service` interface explicitly, even though most methods only delegate to the underlying service without sending events.

Use Go's interface embedding feature to automatically inherit all methods from the `things.Service` interface, then override only the methods that need to send events to Redis.

majabirmancevic:
Resolves #959


bool3max (approved):

manuio (approved):
LGTM

PR #965, opened 2025-12-04
MF-964 - Correct Redis event subscriber lifecycle management

### LINKED ISSUE #964: Fix pkg/events redis subscriber goroutine behavior

bool3max:
Fixes the Redis events subscriber's `.Subscribe()` method to function in a blocking manner and properly react to context cancellation instead of silently spawning a goroutine and exiting.

### DIFF
--- a/pkg/events/subscriber.go
+++ b/pkg/events/subscriber.go
@@ -5,10 +5,10 @@ package events
 
 import (
 	"context"
-	"errors"
 	"fmt"
 
 	"github.com/MainfluxLabs/mainflux/logger"
+	"github.com/MainfluxLabs/mainflux/pkg/errors"
 	"github.com/go-redis/redis/v8"
 )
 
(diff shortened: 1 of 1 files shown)


bool3max:
Once this is merged, services in the enterprise repo which create event subscribers will have to be updated to match this new API. There are no event subscribers in this (main) repo so no changes there.

pkg/events/subscriber.go
```diff
@@ -66,33 +66,37 @@ func NewSubscriber(url, stream, group, consumer string, logger logger.Logger) (S
 }
 
 func (es *subEventStore) Subscribe(ctx context.Context, handler EventHandler) error {
+	defer es.client.Close()
```
majabirmancevic: We have this method
```
func (es *subEventStore) Close() error {
	return es.client.Close()
}
```

so, we could call it from the consumer's main, something like:
```
func subscribeToThingsES(ctx context.Context, svc Service, cfg config, logger logger.Logger) error {
    subscriber, err := events.NewSubscriber(cfg.esURL, ...)
    if err != nil {
        return err
    }
    
    defer func() {
        if err := subscriber.Close(); err != nil {
            logger.Warn(fmt.Sprintf("failed to close Redis subscriber: %v", err))
        }
    }()

...
```
or similar

bool3max: Yup, that's better. I'll remove the `close()` from the subscribe routine.

pkg/events/subscriber.go
```diff
@@ -66,33 +66,37 @@ func NewSubscriber(url, stream, group, consumer string, logger logger.Logger) (S
...
-			}).Result()
-			if err != nil {
-				es.logger.Warn(fmt.Sprintf("failed to read from Redis stream: %s", err))
-
-				continue
-			}
-			if len(msgs) == 0 {
-				continue
+	for {
+		msgs, err := es.client.XReadGroup(ctx, &redis.XReadGroupArgs{
```
majabirmancevic, line 77: We could consider adding a `Block` arg: https://redis.io/docs/latest/commands/xread/#blocking-for-data

bool3max: Sure but I'd leave that for a future PR

majabirmancevic (approved):

manuio (approved):
LGTM

PR #968, opened 2025-12-09
MF-912 - Support optional Group assignment when creating Org Invites

### LINKED ISSUE #912: Add optional Group membership assignment to Org Invites
Support coupling an Org Invite to a specific Group within that Org, so that the invitee is automatically assigned as a member of that Group upon accepting the invite.

bool3max:
* When creating an Org Invite, it's now possible to optionally pair it with one or more Groups (and associated roles) within that Org. Upon accepting the Org Invite, the invitee is then automatically assigned as a member of all the associated Groups
    * The group membership information is supplied in a `"groups"` property of the JSON object of the request for creating an Org Invite -- see the OpenAPI schema for the exact request format
* When creating a 'standard' Group Membership (where the user is already a member of the Group's belonging Org), the recipient now receives an e-mail notification


bool3max:
Groups and Orgs living in different services continues to be a bit of a headache. 

In this PR in particular, when an Org invite associated with one or more groups is accepted by the invitee (which happens in the `auth` service), we need to perform a gRPC request to `things` to create the appropriate group memebrships. The existing service method for this, `CreateGroupMemberships` however, requires authentication (i.e. a `token` value), which is an issue as the invitee making the request to accept the invite obviously isn't a member of the group(s) yet and so we can't supply their token.

I solved this by creating a new [`SaveGroupMemberships`](https://github.com/MainfluxLabs/mainflux/pull/968/files#diff-0a23de2cd1a89314837dc94f67e2c952ac377c581f0ec66eead8721517248f1aR127) service method in `things`, which doesn't require authentication, and is only exposed by the gRPC layer. However I'm iffy about this approach as **we then have two service methods which do pretty much the same thing.**

Now, none of this would be an issue if gRPC endpoint handlers had direct access to the database (or the `<Entity>Repository` type), but that's not currently the case, the endpoint handler only has access to the `things.Service` interface value and so can only access the methods it exposes.

---

A similar issue is present in the `auth` service `ActivateDormantOrgInvites` method, [here](https://github.com/MainfluxLabs/mainflux/pull/968/files#diff-67063a3b5cd7ee93d7a1afaf095934dfbba248b5511d16185c85a47c1e44f92aR295): when a dormant Org invite that's associated with one or more groups is activated, an e-mail notification is sent to the recipient, and for the content of the e-mail to include a list of groups names and roles, we have to fetch human-readable names of the Groups associated with the invite using the `things` service `ViewGroup` method, because an Org Invite stores alongside itself only Group IDs, but not their names. As `ViewGroup` requires authentication though, the same issue reappears here, and again, one solution would be to create a service method similar to `ViewGroup`, but which wouldn't require authentication. 

Another solution here would be to store the group name alongside the invite itself in the database, but then we'd be duplicating data across services, which is a whole another issue.

pkg/proto/mfx.proto
```diff
@@ -205,3 +208,36 @@ message ActivateOrgInviteReq {
...
+message ViewOrgReq {
+    string token = 1;
+    string orgID = 2;
+}
+
+message Org {
+    string id = 1;
+    string ownerID = 2;
+    string name = 3;
+    string description = 4;
```
anna02272: Both Org and Group include a description field. Do we need this in the gRPC models? Are there other fields here that are unused as well?

auth/api/grpc/client.go
```diff
@@ -101,6 +102,14 @@ func NewClient(conn *grpc.ClientConn, tracer opentracing.Tracer, timeout time.Du
 			decodeEmptyResponse,
 			emptypb.Empty{},
 		).Endpoint()),
+		viewOrg: kitot.TraceClient(tracer, "view_org")(kitgrpc.NewClient( // initialize viewOrg client
```
anna02272: remove comment

pkg/proto/mfx.proto
bool3max: Not sure. I thought I'd add it for completeness (in case we ever need it in the future), but now I realize that I skipped the "created at" and "updated at" fields, though from what I can see no other protobuf type has those fields. So I think it's fine to leave the description in.

manuio: If it's unused no need to add it

auth/api/grpc/requests.go
```diff
@@ -154,3 +154,20 @@ func (req activateOrgInviteReq) validate() error {
 
 	return nil
 }
+
+type viewOrgReq struct {
+	token string
+	orgID string
```
majabirmancevic: id

auth/invites.go
```diff
@@ -367,6 +372,25 @@ func (svc service) RespondOrgInvite(ctx context.Context, token, inviteID string,
 		if err := svc.memberships.Save(ctx, membership); err != nil {
 			return err
 		}
+
+		// Create one group membership in the things service for each group the invite was associated with
+		if len(invite.Groups) > 0 {
+			grpcReq := &protomfx.CreateGroupMembershipsReq{
```
majabirmancevic: this part could be separated into a small helper

things/groups.go
```diff
@@ -244,12 +256,11 @@ func (ts *thingsService) ViewGroup(ctx context.Context, token, groupID string) (
...
-	gr, err := ts.groups.RetrieveByID(ctx, groupID)
-	if err != nil {
-		return Group{}, err
-	}
+	return ts.viewGroup(ctx, groupID)
+}
 
-	return gr, nil
+func (ts *thingsService) ViewGroupInternal(ctx context.Context, groupID string) (Group, error) {
+	return ts.viewGroup(ctx, groupID)
```
majabirmancevic, line 292: we could use `return ts.groups.RetrieveByID(ctx, groupID)` here and in `ViewGroup`, without  the additional `ts.viewGroup`

auth/invites.go
bool3max: Hmm not sure I agree, creating memberships is central to this method. Maybe converting the `map[string]string` to a slice of `*protomfx.GroupMembership`, but even that is only done here and it's not that many lines.

things/groups.go
bool3max: Even though the "core" of both `ViewGroup` and `ViewGroupInternal` is just like one line (`ts.groups.RetrieveByID(ctx, groupID)`), I think it's a good idea to extract it into a common method (`viewGroup`). Let's say in the future for some reason the logic for viewing a Group changes (or requires some addition or something), this way you only have to change `viewGroup`. Otherwise, you have to remember to change both `ViewGroup` and `ViewGroupInternal`. << that's my thought process

pkg/proto/mfx.proto
```diff
@@ -29,6 +29,8 @@ service ThingsService {
     rpc GetGroupIDByProfileID(ProfileID) returns (GroupID) {}
     rpc GetGroupIDsByOrg(OrgAccessReq) returns (GroupIDs) {}
     rpc GetThingIDsByProfile(ProfileID) returns (ThingIDs) {}
+    rpc CreateGroupMemberships(CreateGroupMembershipsReq) returns (google.protobuf.Empty) {}
+    rpc ViewGroup(ViewGroupReq) returns (Group) {}
```
bool3max: Rename to `GetGroup` to differentiate between "regular" `ViewGroup` service method which requires auth.

things/groups.go
majabirmancevic: I'd introduce it when the need arises (maybe we "never" need it) :)

auth/invites.go
majabirmancevic: > creating memberships is central to this method.

I'd say that in this method it's an optional action; if Groups are not passed, that part of the code is not executed, therefore it's not "central", right?

api/openapi/auth.yml
```diff
@@ -941,6 +941,14 @@ components:
           type: string
           example: "editor"
           description: The user's proposed role in the organization.
+        groups:
+          type: object
+          description: Optional mapping of Group IDs to roles within that group. If present, upon accepting the invite, the recipient is also assigned as a member of the groups.
+          additionalProperties:
+            type: string
+          example:
+            abcd-efgh-1: "viewer"
```
majabirmancevic: let's use a more meaningful example here: `abcd-efgh-1`, `abcd-efgh-2`

auth/invites.go
bool3max: Right central is a bad word, I meant that no other part of code does this

things/groups.go
bool3max: If/when the need arises you're much less likely to skip or forget to change one of the two methods because you'd see a call to `.viewGroup` and just change that one instead.

auth/postgres/invites.go
```diff
@@ -459,6 +490,83 @@ func (ir invitesRepository) syncOrgInviteStateByID(ctx context.Context, inviteID
 	return nil
 }
 
+func (ir invitesRepository) saveOrgInviteGroups(ctx context.Context, tx *sqlx.Tx, invite auth.OrgInvite) error {
+	if len(invite.Groups) == 0 {
```
manuio: This check should be done before, probably in the service layer

things/groups.go
```diff
@@ -77,6 +77,9 @@ type Groups interface {
 	// ViewGroup retrieves data about the group identified by ID.
 	ViewGroup(ctx context.Context, token, id string) (Group, error)
 
+	// ViewGroupInternal retrieves data about the Group identiified by ID, without requiring authentication.
```
manuio: `identiified`

things/emailer/emailer.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	return &emailer{
+		host:  host,
+		agent: agent,
+	}, nil
+}
+
+func (e *emailer) SendGroupMembershipNotification(to []string, orgName, groupName, groupRole string) error {
+
```
manuio: kill the blank line

manuio (changes requested):

auth/postgres/invites.go
bool3max: This isn't an exported Repository method, simply a helper method that saves the invite's associated groups (if there are any). The service never calls this one (it calls `SaveOrgInvite()` instead, which in turn calls this one).

But now that I think of it this check _is_ unnecessary: if `invite.Groups` has no keys `nil` is simply returned either way.

So I can either:

1) remove this check altogether
2) move it out into `invitesRepository.SaveOrgInvite()`, so that we avoid calling this helper if there the invite has no associated groups

bool3max: I removed the check from the helper. It doesn't make a difference either way, if `invite.Groups` is `nil` or has no keys the helper just does no work and returns `nil` anyway.

auth/invites.go
manuio: What I see here is if cases inside if cases. Let's try to avoid it by returning as soon as possible. Maybe a helper could help here but in any case what if we check if groups are empty and then simply return (and continue outside the if case if not)?

bool3max: That's another way, but as we have to update the invite state at the end, it'd end up looking something like:

```go
// No associated groups, return early
if len(invite.Groups) == 0 {
    return svc.invites.UpdateOrgInviteState(ctx, inviteID, newState)
}

// One or more associated groups, save them with the same code snippet as in the above if clause
// ....
// ....


// Sync invite state at the end
return svc.invites.UpdateOrgInviteState(ctx, inviteID, newState)
```

This way there's not as many nested `if`s but we have to repeat the call to `UpdateOrgInviteState` twice.

Here we could defer the call to `UpdateOrgInviteState` somewhere at an earlier point in the method to overcome this but I'm not a fan of `defer` for cases like this.

---

If we do instead define a little helper method for creating the memberships, e.g.:

```go
func (...) createGroupMemerships(userID string, groups map[string]string) error
```

The snippet in the main method would look something like:

```go
if len(invite.Groups) > 0 {
    if err := svc.createGroupMemberships(inviteeID, groups); err != nil {
        return err
    }    
}
```

... so, still nested `if`s. I personally think it's fine the way it is now, it's nested only one level deep and still pretty readable. Extracting the logic into a method would tidy it up even more, but I don't know if it's worth it to define a helper here just for this one occurrence.

things/api/metrics.go
```diff
@@ -507,6 +507,15 @@ func (ms *metricsMiddleware) CreateGroupMemberships(ctx context.Context, token s
 	return ms.svc.CreateGroupMemberships(ctx, token, gms...)
 }
 
+func (ms *metricsMiddleware) CreateGroupMembershipsInternal(ctx context.Context, gms ...things.GroupMembership) error {
+	defer func(begin time.Time) {
+		ms.counter.With("method", "save_group_memberships").Add(1)
```
majabirmancevic: rename `save_group_memberships`

things/api/logging.go
```diff
@@ -714,6 +714,18 @@ func (lm *loggingMiddleware) CreateGroupMemberships(ctx context.Context, token s
 	return lm.svc.CreateGroupMemberships(ctx, token, gms...)
 }
 
+func (lm *loggingMiddleware) CreateGroupMembershipsInternal(ctx context.Context, gms ...things.GroupMembership) (err error) {
+	defer func(begin time.Time) {
+		message := fmt.Sprintf("Method save_group_memberships took %s to complete", time.Since(begin))
```
majabirmancevic: and here

things/api/grpc/server.go
```diff
@@ -22,17 +22,19 @@ import (
...
+	canUserAccessProfile   kitgrpc.Handler
+	canUserAccessGroup     kitgrpc.Handler
+	canThingAccessGroup    kitgrpc.Handler
+	identify               kitgrpc.Handler
+	getGroupIDByThingID    kitgrpc.Handler
+	getGroupIDByProfileID  kitgrpc.Handler
+	getGroupIDsByOrg       kitgrpc.Handler
+	getThingIDsByProfile   kitgrpc.Handler
+	createGroupMemberships kitgrpc.Handler
+	viewGroup              kitgrpc.Handler
```
majabirmancevic: getGroup?

bool3max: It's already `getGroup`, you might be looking at an older version

pkg/proto/mfx.proto
```diff
@@ -45,6 +47,7 @@ service AuthService {
     rpc RetrieveRole(RetrieveRoleReq) returns (RetrieveRoleRes) {}
     rpc CreateDormantOrgInvite(CreateDormantOrgInviteReq) returns (google.protobuf.Empty) {}
     rpc ActivateOrgInvite(ActivateOrgInviteReq) returns (google.protobuf.Empty) {}
+    rpc ViewOrg(ViewOrgReq) returns (Org) {}
```
majabirmancevic, line 50: GetOrg?

bool3max: This one's for the regular ViewOrg service method, with auth.

things/api/grpc/responses.go
```diff
@@ -35,3 +35,10 @@ type groupIDsRes struct {
 type thingIDsRes struct {
 	thingIDs []string
 }
+
+type groupRes struct {
+	id          string
+	orgID       string
+	name        string
+	description string
```
anna02272: Do we need description field?

auth/api/grpc/requests.go
majabirmancevic: ^ resolved?

things/api/grpc/responses.go
bool3max: Ah I forgot to remove it after deleting the field from the protobuf definition

auth/api/grpc/requests.go
bool3max: Not sure why it was marked as resolved, but it should be good now

auth/invites.go
manuio: I agree in your example but there are 3 nested if cases. What if we had spmething like that?
```
	if accept {
		return svc.acceptInvite(ctx, user, invite)
	}
	
	// Decline is simple - just update state
	return svc.invites.UpdateOrgInviteState(ctx, inviteID, InviteStateDeclined)
```

bool3max: That looks good, but then you have to update the invite's state in the `svc.acceptInvite` method and I think it's clearer if only the primary `RespondOrgInvite` is the one managing the state.

To overcome that we could do something like:

```go
newState := InviteStateDeclined

if accept {
    newState = InviteStateAccepted
    if err := svc.acceptInvite(ctx, invite) {
        return err
    }
}


return svc.invites.UpdateOrgInviteState(ctx, inviteID, newState)
```

This way state management is all in the primary method and clearly visible, however we do have one nested `if`, which I don't think is that bad at all. What do you think?

manuio: ok, I agree

things/groups.go
manuio: We can keep it for now

anna02272 (approved):

manuio (approved):
LGTM

manuio:
`func (lm *loggingMiddleware) CreateOrgInvite(ctx context.Context, token, email, role, orgID string, groups map[string]string, invRedirectPath string) (invite auth.OrgInvite, err error)`

@bool3max `groups` argument should be a list of `GroupMembership` instead of a `map[string]string`. Change it in your next PR.

PR #986, opened 2025-12-19
MF-985 - Publish messages to NATS from adapters

### LINKED ISSUE #985: Publish messages to NATS from adapters
Currently, incoming messages are forwarded to the rules service, which publishes them to default topics (messages, writers) + to topics that depend on defined actions (smtp, smpp, alarms).
Instead, return message publishing to default NATS topics directly from the adapters and enable the rules service to subscribe to a specific topic (e.g., messages) and republish further to other topics (smtp, smpp, alarms).

majabirmancevic:
Resolves #985


manuio (approved):
LGTM

duolok (approved):

Issue #987, opened 2025-12-19
Rename things and auth gRPC methods


majabirmancevic:
Rename:
GetConfigByThingID -> GetConfigByThing
GetGroupIDByThingID -> GetGroupIDByThing
GetGroupIDByProfileID -> GetGroupIDByProfile
GetOwnerIDByOrgID -> GetOwnerIDByOrg
GetPubConfByKey -> GetPubConfigByKey

PR #989, opened 2025-12-22
MF-988 - Refactor MQTT subscriptions to be thing-centric

### LINKED ISSUE #988: Refactor MQTT subscriptions to be thing-centric
The MQTT adapter previously treated subscriptions as client-specific entities. This resulted in several issues:
- Multiple identical subscriptions could be created for the same thing when multiple MQTT clients connected using the same credentials
- Subscription lifecycle was tightly coupled to transient MQTT client connections
- `identify` was called multiple times per client lifecycle, causing unnecessary load on the `Things` service
- Connection state (`status`) was persisted in the database even though it is inherently runtime-only

majabirmancevic:
This change simplifies MQTT subscription handling and reduces runtime overhead by separating _persistent configuration_ from _connection state_.

Subscriptions are now thing-centric
- Subscriptions are uniquely identified by (subtopic, group_id, thing_id)
- `client_id` is removed from the primary key and the table
- A subscription represents a thing’s intent to receive messages, independent of how many MQTT clients are connected

_Multiple clients connecting as the same thing should not create duplicate subscriptions or cause conflicts._

---
Duplicate subscriptions are treated as a no-op
- Existing duplicate rows are deduplicated during migration (`mqtt_3`)
- The oldest subscription (by created_at) is preserved
- Creating an already-existing subscription is not an error

_Repeated SUBSCRIBE calls are valid in MQTT and should be idempotent._

---
Runtime connection state removed from the database
- `status` column is removed
- Connection state is handled at runtime only
- Database stores configuration, not live connection data

_Connection state is transient and does not belong in persistent storage._

---
Redis cache added for client → thing mapping
- Cache stores: `client_id` → `thing_id`
- Used to avoid repeated `identify` calls across MQTT hooks

_Improves performance and reduces load on the Things service._

---
MQTT hook behavior clarified
- AuthConnect: identifies the thing and caches the mapping
- Other hooks reuse cached data when available
- Disconnect clears cache only; subscriptions are unaffected

_Subscriptions are thing-scoped and should not depend on client lifecycle._

Resolves #988


mqtt/handler.go
```diff
@@ -260,13 +262,13 @@ func (h *handler) getSubscriptions(c *session.Client, topics *[]string) ([]Subsc
 		return nil, err
 	}
 
+	groupID, err := h.things.GetGroupIDByThingID(context.Background(), &protomfx.ThingID{Value: thingID})
```
manuio, line 265: `GetGroupByThing` or rename `GetThingByClient`

mqtt/redis/subscriptions.go
```diff
@@ -0,0 +1,64 @@
...
+
+	val, err := m.client.Get(ctx, key).Result()
+	if err != nil {
+		return ""
+	}
+
+	return val
+}
+
+func thingByClientKey(clientID string) string {
```
manuio: Let's add a prefix to make it more understandable

manuio (changes requested):

mqtt/handler.go
majabirmancevic: GetGroupIDByThingID should be renamed in another PR (https://github.com/MainfluxLabs/mainflux/issues/987),
while `GetThingByClient` can be `ThingByClient`?

mqtt/redis/subscriptions.go
majabirmancevic: the name is more or less consistent with the existing helpers: https://github.com/MainfluxLabs/mainflux/blob/525e890db445e186c7ec65761dcc969382d4cafd/things/redis/things.go#L179
maybe `thingByClientIDKey` instead

mqtt/handler.go
manuio: Or `RetrieveThingByClient`? Why to remove the prefix?

manuio (approved):
LGTM

anna02272 (approved):

PR #1004, opened 2026-01-26
NOISSUE - Add gRPC methods for retrieving thing IDs by group/org

anna02272:
This PR adds two new gRPC methods to the Things service: `GetThingIDsByGroup` and `GetThingIDsByOrg`.

This is needed because the retrieved IDs are used to delete UI configs when group or org removal events are received.


things/service.go
```diff
@@ -1146,3 +1152,34 @@ func (ts *thingsService) GetThingIDsByProfile(ctx context.Context, profileID str
...
+func (ts *thingsService) GetThingIDsByGroup(ctx context.Context, groupID string) ([]string, error) {
+	page, err := ts.things.RetrieveByGroups(ctx, []string{groupID}, apiutil.PageMetadata{})
+	if err != nil {
+		return []string{}, err
+	}
+
+	var thingIDs []string
+	for _, t := range page.Things {
+		thingIDs = append(thingIDs, t.ID)
+	}
```
bool3max: Whenever you know the final size of the slice in advance like this, you can preallocate its capacity:

```go
thingIDs := make([]string, 0, len(page.Things))
...
```

things/service.go
```diff
@@ -1146,3 +1152,34 @@ func (ts *thingsService) GetThingIDsByProfile(ctx context.Context, profileID str
...
+
+	page, err := ts.things.RetrieveByGroups(ctx, grIDs, apiutil.PageMetadata{})
+	if err != nil {
+		return []string{}, err
+	}
+
+	var thingIDs []string
+	for _, t := range page.Things {
+		thingIDs = append(thingIDs, t.ID)
+	}
```
bool3max: Same as above

things/api/grpc/requests.go
```diff
@@ -122,6 +122,30 @@ func (req profileIDReq) validate() error {
 	return nil
 }
 
+type groupIDReq struct {
```
majabirmancevic: this structure already exists; we can rename it and use it where needed instead of creating a new one
```
type getGroupReq struct {
	groupID string
}
```

Issue #995, opened 2026-01-08
Refactor backups endpoints


manuio:
Remove partial backup and restore endoints:
```
	mux.Get("/groups/:id/things/backup", kithttp.NewServer(
		kitot.TraceServer(tracer, "backup_things_by_group")(backupThingsByGroupEndpoint(svc)),
		decodeBackupThingsByGroup,
		apiutil.EncodeFileResponse,
		opts...,
	))

	mux.Post("/groups/:id/things/restore", kithttp.NewServer(
		kitot.TraceServer(tracer, "restore_things_by_group")(restoreThingsByGroupEndpoint(svc)),
		decodeRestoreThingsByGroup,
		encodeResponse,
		opts...,
	))

	mux.Get("/orgs/:id/things/backup", kithttp.NewServer(
		kitot.TraceServer(tracer, "backup_things_by_org")(backupThingsByOrgEndpoint(svc)),
		decodeBackupThingsByOrg,
		apiutil.EncodeFileResponse,
		opts...,
	))

	mux.Post("/orgs/:id/things/restore", kithttp.NewServer(
		kitot.TraceServer(tracer, "restore_things_by_org")(restoreThingsByOrgEndpoint(svc)),
		decodeRestoreThingsByOrg,
		encodeResponse,
		opts...,
	))
```

Create separate directory for service backups.
Use correct naming for alarms and messages reports (instead of backup)

PR #1011, opened 2026-01-30
MF-995 - Merge JSON and SenML backup into single endpoint and reorganize directories

anna02272:
Related to #995 

- Combined JSON and SenML backup endpoints into a single `/backup` endpoint.
- Removed separate backup endpoints for JSON and SenML messages.
- Renamed backup functionality to "export" to reflect usage for report generation.
- Reorganized directories


readers/service.go
```diff
@@ -22,17 +28,17 @@ type Service interface {
...
 	ListSenMLMessages(ctx context.Context, token string, key things.ThingKey, rpm SenMLPageMetadata) (SenMLMessagesPage, error)
 
-	// BackupJSONMessages backups the json messages with given filters.
-	BackupJSONMessages(ctx context.Context, token string, rpm JSONPageMetadata) (JSONMessagesPage, error)
+	// ExportJSONMessages retrieves the json messages with given filters, intended for exporting.
+	ExportJSONMessages(ctx context.Context, token string, rpm JSONPageMetadata) (JSONMessagesPage, error)
 
-	// BackupSenMLMessages backups the senml messages with given filters.
-	BackupSenMLMessages(ctx context.Context, token string, rpm SenMLPageMetadata) (SenMLMessagesPage, error)
+	// ExportSenMLMessages retrieved the senml messages with given filters, intended for exporting.
```
duolok: retrieves

duolok (changes requested):

manuio (dismissed):
LGTM

api/openapi/readers.yml
```diff
@@ -185,14 +185,14 @@ paths:
...
+  /json/export:
     get:
-      summary: Backup JSON messages
+      summary: Export JSON messages
       security:
         - bearerAuth: []
       description: |
-        Creates a backup of JSON messages with optional filtering and format conversion.
-      tags: 
+        Creates an export of JSON messages with optional filtering and format conversion.
```
majabirmancevic: Exports JSON messages ... or Creates file ...

readers/api/http/backup/endpoint.go
```diff
@@ -0,0 +1,91 @@
...
+			JSONMessages:  json,
+			SenMLMessages: senml,
+		}, nil
+	}
+}
+
+func restoreEndpoint(svc readers.Service) endpoint.Endpoint {
+	return func(ctx context.Context, request any) (any, error) {
+		req := request.(restoreReq)
+
```
majabirmancevic: remove blank line

readers/api/http/backup/requests.go
```diff
@@ -0,0 +1,39 @@
...
+func (req backupReq) validate() error {
+	if req.token == "" {
+		return apiutil.ErrBearerToken
+	}
+	return nil
+}
+
+type restoreReq struct {
+	token         string
+	JSONMessages  json.RawMessage `json:"json_messages"`
```
majabirmancevic, line 25: not sure we use `json.RawMessage` anywhere, mostly we use []byte and map?

manuio: Right. Plus, it's also used for senml messages.

anna02272: I used `json.RawMessage` because our existing `converter` functions work with []byte. Plain []byte isn’t human-readable, but json.RawMessage preserves the JSON format while still allowing us to reuse our converters.

majabirmancevic: The advantage may be that it keeps the JSON format instead of encoding in base64 like with plain []byte;
but this `JSONMessages  json.RawMessage json:"json_messages"`
may seem confusing since the field name is plural and the type is "singular", although json.RawMessage supports an array of JSON objects?

anna02272: I understand it may look confusing because the type name is singular, but json.RawMessage can store any valid JSON.

Previously we returned a single file as raw bytes per endpoint. But now we return a single JSON response that contains both JSON and SenML messages. 

Using map[string]any would work, but we wouldn’t be able to reuse the existing converter functions without extra marshaling/unmarshaling, which would complicate the code. 

Using json.RawMessage makes the messages readable and works with our existing converters...

majabirmancevic (approved):

manuio (approved):
LGTM

duolok (approved):

PR #1018, opened 2026-02-05
MF-993 - Refactor mqtt Publish method

### LINKED ISSUE #993: Consider message formats for MQTT publishing
Currently, MQTT messages are serialized using Protobuf ([check here](https://github.com/MainfluxLabs/mainflux/blob/5f033102e53a1e6c42c7bdcc069b2296effca485/pkg/messaging/mqtt/publisher.go#L60)) when published, which complicates integration with lightweight devices and simple MQTT clients that expect JSON.
We should consider using Protobuf only within the service (and for internal communication, of course), making it optional for external communication, while using a more interoperable format like JSON by default.

majabirmancevic:
Resolves #993


manuio (approved):
LGTM

duolok (approved):
LGTM

anna02272 (approved):

PR #1024, opened 2026-02-09
MF-1014 - Expand list of allowed columns in dbutil.GetOrderQuery

### LINKED ISSUE #1014: Expand column support in dbutil.GetOrderQuery
`dbutil.GetOrderQuery` which is relied upon when building queries for fetching many types of entities currently allows ordering on two columns only:

https://github.com/MainfluxLabs/mainflux/blob/aa5872d4240ca0955caf6dd519d060e6fe5e85a5/pkg/dbutil/query.go#L54-L63

... which leads to broken sorting order from API endpoints for fetching lists of various kinds of entities.

Add checks for other common column names (`"created"`, `"created_at"`, `"updated_at"`, etc...)

bool3max:
Resolves #1014


manuio (approved):
LGTM

majabirmancevic (approved):

PR #1025, opened 2026-02-09
MF-1017 - Properly validate invitee Org role when creating Platform Invite

### LINKED ISSUE #1017: Disallow "owner" destination Org role when creating Platform+Org invites

bool3max:
Resolves #1017

### DIFF
--- a/auth/service.go
+++ b/auth/service.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"time"
 
+	"github.com/MainfluxLabs/mainflux/pkg/apiutil"
 	"github.com/MainfluxLabs/mainflux/pkg/errors"
 	protomfx "github.com/MainfluxLabs/mainflux/pkg/proto"
 	"github.com/MainfluxLabs/mainflux/pkg/uuid"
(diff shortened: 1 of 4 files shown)


auth/service.go
```diff
@@ -234,3 +235,11 @@ func (svc service) isAdmin(ctx context.Context, token string) error {
 
 	return nil
 }
+
+func ValidateRole(role string) error {
+	if role != Owner && role != Admin && role != Editor && role != Viewer {
```
majabirmancevic: do we need to validate the Owner role at all?

`if role != Admin && role != Editor && role != Viewer ` could be enough since we don't allow assigning the `owner` role?

majabirmancevic: and to be more precise, we can name the helper `ValidateInviteeRole`

bool3max: I've been thinking of this method as a more general one that validates any `auth`-related role, but seeing as it's not really used anywhere besides validating roles of invitees, yeah, we can ditch the owner check and use it like that

users/api/http/invites/requests.go
```diff
@@ -79,8 +79,10 @@ func (req createPlatformInviteRequest) validate() error {
 		return apiutil.ErrMissingRedirectPath
 	}
 
-	if req.OrgID != "" && req.Role == "" {
-		return apiutil.ErrMissingRole
+	if req.OrgID != "" {
```
manuio, line 82: Return a specific error if org ID is empty

manuio (changes requested):

users/api/http/invites/requests.go
bool3max: `OrgID` is optional here (if specified a dormant Org invite is created and tied to the platform invite)

majabirmancevic (approved):

users/api/http/invites/requests.go
manuio: oh, my bad

manuio (approved):
LGTM

PR #1089, opened 2026-02-26
MF-1088 -  Use consistent URLs for Redis cache and event store

### LINKED ISSUE #1088: Use consistent URLs for Redis cache and event store
Switch cache client initialization from:
```go
redis.NewClient(&redis.Options{Addr: cacheURL})
```
to:
```go
opts, _ := redis.ParseURL(cacheURL)
redis.NewClient(opts)
```

This allows all Redis URLs (both cache and event store) to use the standard `redis://host:port/0` format consistently.

majabirmancevic:
Resolves #1088


majabirmancevic:
Consider moving `connectToRedis` function to `pkg` in the new PR

cmd/auth/main.go
```diff
@@ -69,9 +69,7 @@ const (
 	defUsersCACerts    = ""
 	defUsersClientTLS  = "false"
 	defUsersGRPCURL    = "localhost:8184"
-	defESURL           = "localhost:6379"
-	defESPass          = ""
```
manuio, line 73: Is it possible to have a Redis password for events or it's only for the cache?

majabirmancevic: Yes, it's possible to configure a password within the URL for both:
`redis://[:password@]host[:port][/db-number]`
In our case, the password was always empty and DB was always 0, but if we ever need a password, we just put it in the URL: `redis://:mypassword@localhost:6379/0` (no code change needed, just a config change).

anna02272 (approved):

manuio (approved):
LGTM

PR #1098, opened 2026-03-02
MF-1097 - Authorize subscription topics for commands and messages

### LINKED ISSUE #1097: Authorize subscription topics for commands and messages
Add authorization when devices subscribe to MQTT topics so a device can only subscribe to topics that belong to it.

- **Thing commands** (`things/{id}/commands/...`): subscription is allowed only if the topic’s thing ID is the same as the authenticated thing.
- **Group commands** (`groups/{id}/commands/...`): subscription is allowed only if the topic’s group ID is the same as the thing’s group (from `GetGroupIDByThing`).
- **Thing messages** (`things/{id}/messages/...`): subscription is allowed only if the topic’s thing ID is the same as the authenticated thing.

Topics that do not match these patterns (e.g. custom or third‑party topics) are not checked and are still allowed.

majabirmancevic:
Resolves #1097


anna02272 (dismissed):

mqtt/handler.go
```diff
@@ -282,3 +303,50 @@ func (h *handler) getSubscriptions(c *session.Client, topics *[]string) ([]Subsc
 
 	return subs, nil
 }
+
+// checkCustomTopic enforces authorization only for topics that match
+// custom patterns (things/thingID/commands, groups/groupID/commands, things/thingID/messages).
+func checkCustomTopic(topic, thingID, groupID string) error {
```
manuio: I'd call it `validateTopic`

manuio (changes requested):

anna02272 (approved):

manuio (approved):
LGTM

PR #1111, opened 2026-03-05
MF-1112 - Add endpoint for listing API Keys

### LINKED ISSUE #1112: Allow users to list their API keys
Users can currently issue and revoke API keys, but cannot list existing keys, making key management harder.
We should add a paginated GET /keys endpoint.

anna02272:
Adds a new authenticated endpoint `GET /keys `to list API keys for the current user.

Resolves #1112


duolok (dismissed):
LGTM

auth/api/http/keys/responses.go
```diff
@@ -55,6 +55,31 @@ func (res retrieveKeyRes) Empty() bool {
...
+type keysPageRes struct {
+	pageRes
+	Keys []retrieveKeyRes `json:"keys"`
+}
+
+type pageRes struct {
+	Limit  uint64 `json:"limit"`
+	Offset uint64 `json:"offset"`
+	Total  uint64 `json:"total"`
+	Ord    string `json:"order,omitempty"`
```
manuio: Use same naming for the field and the json tag

auth/api/http/keys/responses.go
```diff
@@ -55,6 +55,31 @@ func (res retrieveKeyRes) Empty() bool {
...
+	Dir    string `json:"direction,omitempty"`
```
manuio: Aren't we using `dir` in other services?

manuio (changes requested):

manuio (approved):
LGTM

duolok (approved):

majabirmancevic (approved):

PR #1114, opened 2026-03-06
MF-1156 - refactor: extract shared domain types to domain subpackages

### LINKED ISSUE #1156: Extract shared service domain types to separate shared package
As a precursor to resolving #1096 it's necessary to break any existing inter-service imports. That is, top-level service packages should not ever depend on one another.

To ensure that services can seamlessly call one another's gRPC client methods in a clean manner using service-domain types (e.g. `auth.AuthzReq`, `users.User`, etc...), these types must not belong to the service packages themselves.

bool3max:
Shared domain types from top-level services have been extracted to a shared `/pkg/domain/` package, with domain types grouped per service, each one having its own file.

To minimize the number of changes in the service packages themselves, type aliases which reference the new domain types are defined. For example, `auth` defines:

```go
type OrgInvite = domain.OrgInvite
```

... so it can continue to reference `OrgInvite` in its code. Note that `OrgInvite` is a type alias and not a distinct type.

---

Even disregarding the upcoming gRPC client interface refactor, this PR resolves many existing inter-service imports.


auth/keys.go
```diff
@@ -23,44 +23,23 @@ var (
...
+type Key = domainauth.Key
 
-// Identity contains ID and Email.
-type Identity struct {
-	ID    string
-	Email string
-}
+type Identity = domainauth.Identity
 
 type KeysPage struct {
```
anna02272: This should be in domain then also?

bool3max: Yes! Good catch, I didn't notice it after merging changes from master

auth/api/http/invites/endpoint.go
```diff
@@ -15,10 +15,10 @@ func createOrgInviteEndpoint(svc auth.Service) endpoint.Endpoint {
 		}
 
 		orgInvite := auth.OrgInvite{
-			InviteeEmail: req.Email,
 			InviteeRole:  req.Role,
 			OrgID:        req.orgID,
 			GroupInvites: req.GroupInvites,
+			InviteeEmail: req.Email,
```
manuio: Is there a reason to change the field order?

auth/api/http/invites/endpoint_test.go
```diff
@@ -224,9 +224,9 @@ func TestViewInvite(t *testing.T) {
 	assert.Nil(t, err, fmt.Sprintf("Creating Org expected to succeed: %s", err))
 
 	invite, err := svc.CreateOrgInvite(context.Background(), ownerToken, auth.OrgInvite{
-		InviteeEmail: viewer.Email,
 		InviteeRole:  viewer.Role,
 		OrgID:        org.ID,
+		InviteeEmail: viewer.Email,
```
manuio: And here in all other palces?

manuio (changes requested):

auth/api/http/invites/endpoint_test.go
bool3max: That's an LLMs work, not sure why it shuffled the order of the struct fields. I'll try to revert them back manually.

users/postgres/invites.go
```diff
@@ -156,12 +156,8 @@ func (ir invitesRepository) RetrievePlatformInvites(ctx context.Context, pm user
 	}
 
 	page := users.PlatformInvitesPage{
+		Total:   total,
 		Invites: invites,
-		PageMetadata: apiutil.PageMetadata{
```
anna02272, line 160: Is this correct?

auth/mocks/users.go
```diff
@@ -15,11 +15,11 @@ import (
 var _ protomfx.UsersServiceClient = (*usersServiceClientMock)(nil)
 
 type usersServiceClientMock struct {
-	usersByID     map[string]users.User
-	usersByEmails map[string]users.User
+	usersByID     map[string]domainusers.User
```
majabirmancevic: can we use `domain` as an alias to avoid redundancy `domainusers.User`?

users/postgres/invites.go
bool3max: Yeah

auth/memberships.go
```diff
@@ -19,21 +22,8 @@ var (
...
-}
-
-// OrgMembershipsPage contains page related metadata as well as list of memberships that
-// belong to this page.
-type OrgMembershipsPage struct {
-	Total          uint64
-	OrgMemberships []OrgMembership
-}
+// OrgMembershipsPage is an alias for the shared domain type.
+type OrgMembershipsPage = domainauth.OrgMembershipsPage
```
manuio: Should we define all aliases at the top of the file? Check all other pelase

manuio (changes requested):

auth/service.go
```diff
@@ -153,9 +151,9 @@ func New(orgs OrgRepository, tc protomfx.ThingsServiceClient, uc protomfx.UsersS
 
 func (svc service) Authorize(ctx context.Context, ar AuthzReq) error {
 	switch ar.Subject {
-	case RootSub:
+	case domain.RootSub:
```
majabirmancevic: do we use these consts or directly from the pkg domain?
```
	RootSub = domain.RootSub
	OrgSub  = domain.OrgSub
```

do we actually need to re-export constants?

pkg/apiutil/token.go
```diff
@@ -27,3 +29,44 @@ func ExtractBearerToken(r *http.Request) string {
...
+			Value: strings.TrimPrefix(header, ThingKeyPrefixExternal),
+		}
+	}
+
+	return domain.ThingKey{}
+}
+
+// ValidateInviteeRole returns an error if the role is not a valid invitee role
+// (OrgAdmin, OrgEditor, or OrgViewer). OrgOwner cannot be assigned via invite.
+func ValidateInviteeRole(role string) error {
```
majabirmancevic: I think we should find a better place for this helper `ValidateInviteeRole`
and also to think about the name of this file `token.go` since we have thing key stuff

pkg/apiutil/token.go
```diff
@@ -27,3 +29,44 @@ func ExtractBearerToken(r *http.Request) string {
 
 	return strings.TrimPrefix(token, BearerPrefix)
 }
+
+// ExtractThingKeyFromHTTPHeader returns the thing key and its type from the request's HTTP 'Authorization' header.
+// If the provided key type is invalid, an empty ThingKey is returned.
+func ExtractThingKeyFromHTTPHeader(r *http.Request) domain.ThingKey {
```
majabirmancevic: maybe we could consider adding a subdir to avoid `FromHTTPHeader` suffix; let's say apiutil/http/<file.go> and then `http.ExtractThingKey`?

auth/service.go
bool3max: We don't _need_ to re-export them, it was done to keep the number of diffs to a minimum, but these two lines somehow snuck in and reference the domain instead, I'll change them

pkg/apiutil/token.go
bool3max: Yeah, maybe both `ExtractBearerToken` and `ExtractThingKeyFromHTTPHeader` could go into e.g. `apiutil/auth.go` and the two validation functions (`ValidateInviteeRole` and `ValidateThingKey`) could go into `apiutil/validation.go` or something

pkg/apiutil/token.go
bool3max: I think that'd be a bit overkill

manuio: IMO `ExtractThingKey` in apiutil package it's ok

majabirmancevic: ^ the separate files you suggested would be enough for now

pkg/apiutil/token.go
manuio: or `request.go`

pkg/domain/auth.go
```diff
@@ -0,0 +1,136 @@
...
+	RootSub = "root"
+	OrgSub  = "org"
+)
+
+// Org role constants.
+const (
+	OrgAdmin  = "admin"
+	OrgOwner  = "owner"
+	OrgEditor = "editor"
+	OrgViewer = "viewer"
```
majabirmancevic: or should we unify them?

```
// Group role constants.
const (
	GroupViewer = "viewer"
	GroupEditor = "editor"
	GroupAdmin  = "admin"
	GroupOwner  = "owner"
)
```

consumers/alarms/service.go
```diff
@@ -90,7 +90,7 @@ func New(things protomfx.ThingsServiceClient, alarms AlarmRepository, idp uuid.I
 }
 
 func (as *alarmService) ListAlarmsByGroup(ctx context.Context, token, groupID string, pm PageMetadata) (AlarmsPage, error) {
-	_, err := as.things.CanUserAccessGroup(ctx, &protomfx.UserAccessReq{Token: token, Id: groupID, Action: things.Viewer})
+	_, err := as.things.CanUserAccessGroup(ctx, &protomfx.UserAccessReq{Token: token, Id: groupID, Action: domain.GroupViewer})
```
majabirmancevic, line 93: domain.GroupViewer or things.Viewer (because to re-export?)

ws/api/endpoints.go
```diff
@@ -49,7 +49,7 @@ func handshake(svc ws.Service) http.HandlerFunc {
 }
 
 func decodeRequest(r *http.Request) (getConnByKey, error) {
-	authKey := apiutil.ExtractThingKeyFromHTTPHeader(r)
+	authKey := apiutil.ExtractThingKey(r)
```
majabirmancevic: authKey -> thingKey

pkg/domain/auth.go
bool3max: I don't think so, they're roles for different kinds of entities. I think having separate consts is much preferred, despite the actual values coincidentally being the same.

consumers/alarms/service.go
bool3max: `domain.GroupViewer` - this is part of the reason the domain package was introduced, to get rid of cross-service imports. We don't want `consumers/alarms` to depend on `things`.

pkg/domain/things.go
```diff
@@ -0,0 +1,146 @@
+// Copyright (c) Mainflux
+// SPDX-License-Identifier: Apache-2.0
+
+package domain
+
+import (
```
majabirmancevic, line 6: inline

pkg/domain/auth.go
```diff
@@ -0,0 +1,136 @@
+// Copyright (c) Mainflux
+// SPDX-License-Identifier: Apache-2.0
+
+package domain
+
+import (
+	"time"
```
majabirmancevic, line 7: inline

pkg/domain/auth.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// GroupInvite represents a group membership invite.
+type GroupInvite struct {
+	GroupID    string `json:"group_id,omitempty"`
+	MemberRole string `json:"member_role,omitempty"`
+}
+
+// Subject type constants for authorization.
+const (
```
majabirmancevic: maybe it would be clearer to have consts and vars at the top of the files

readers/service.go
```diff
@@ -69,10 +68,10 @@ func New(auth protomfx.AuthServiceClient, things protomfx.ThingsServiceClient, j
 	}
 }
 
-func (rs *readersService) ListJSONMessages(ctx context.Context, token string, key things.ThingKey, rpm JSONPageMetadata) (JSONMessagesPage, error) {
+func (rs *readersService) ListJSONMessages(ctx context.Context, token string, key domain.ThingKey, rpm JSONPageMetadata) (JSONMessagesPage, error) {
 	switch {
 	case rpm.Publisher != "":
-		_, err := rs.thingc.CanUserAccessThing(ctx, &protomfx.UserAccessReq{Token: token, Id: rpm.Publisher, Action: auth.Viewer})
+		_, err := rs.thingc.CanUserAccessThing(ctx, &protomfx.UserAccessReq{Token: token, Id: rpm.Publisher, Action: domain.OrgViewer})
```
majabirmancevic: or `domain.Group<Role>`?

pkg/domain/users.go
```diff
@@ -0,0 +1,64 @@
...
+type User struct {
+	ID       string   `json:"id,omitempty"`
+	Email    string   `json:"email,omitempty"`
+	Password string   `json:"password,omitempty"`
+	Metadata Metadata `json:"metadata,omitempty"`
+	Status   string   `json:"status,omitempty"`
+	Role     string   `json:"role,omitempty"`
+}
+
+// ValidateUser returns an error if user representation is invalid.
```
majabirmancevic: Validate

things/things.go
```diff
@@ -5,39 +5,24 @@ package things
...
-type ThingsPage struct {
-	Total  uint64
-	Things []Thing
-}
+// Thing and ThingsPage are aliases for the shared domain types.
+type Thing = domain.Thing
+type ThingsPage = domain.ThingsPage
+
+// ThingKey is an alias for the shared domain type
+type ThingKey = domain.ThingKey
```
duolok: Put all type aliases inside a single `type()`. Possibly remove duplicated comments.

readers/service.go
```diff
@@ -6,10 +6,9 @@ package readers
...
 	"context"
 
-	"github.com/MainfluxLabs/mainflux/auth"
 	"github.com/MainfluxLabs/mainflux/pkg/apiutil"
+	"github.com/MainfluxLabs/mainflux/pkg/domain"
 	protomfx "github.com/MainfluxLabs/mainflux/pkg/proto"
-	"github.com/MainfluxLabs/mainflux/things"
 )
 
 const rootSubject = "root"
```
majabirmancevic: should we keep this?

users/users.go
```diff
@@ -5,27 +5,16 @@ package users
...
-	Password string
-	Metadata Metadata
-	Status   string
-	Role     string
-}
+// Metadata is an alias for the shared domain type.
+type Metadata = domain.Metadata
+
+// User is an alias for the shared domain type.
+type User = domain.User
```
duolok: Also here

auth/orgs.go
```diff
@@ -12,32 +12,10 @@ var (
...
-}
+// Org is an alias for the shared domain type.
+type Org = domain.Org
 
-type User struct {
-	ID     string
-	Email  string
-	Status string
-}
+type OrgsPage = domain.OrgsPage
```
duolok: here

auth/keys.go
```diff
@@ -22,43 +22,17 @@ var (
...
-func (k Key) Expired() bool {
-	if k.Type == APIKey && k.ExpiresAt.IsZero() {
-		return false
-	}
-	return k.ExpiresAt.UTC().Before(time.Now().UTC())
-}
+// Key is an alias for the shared domain type.
+type Key = domain.Key
+type Identity = domain.Identity
+type KeysPage = domain.KeysPage
```
duolok: here

auth/invites.go
```diff
@@ -16,30 +17,14 @@ import (
...
-}
+// OrgInvitesPage is an alias for the shared domain type.
+type OrgInvitesPage = domain.OrgInvitesPage
 
-type GroupInvite struct {
-	GroupID    string `json:"group_id"`
-	MemberRole string `json:"member_role"`
-}
+// GroupInvite is an alias for the shared domain type.
+type GroupInvite = domain.GroupInvite
```
duolok: here

auth/memberships.go
```diff
@@ -2,12 +2,18 @@ package auth
...
 	"github.com/MainfluxLabs/mainflux/pkg/errors"
 	protomfx "github.com/MainfluxLabs/mainflux/pkg/proto"
 )
 
+// OrgMembership is an alias for the shared domain type.
+type OrgMembership = domain.OrgMembership
+
+// OrgMembershipsPage is an alias for the shared domain type.
+type OrgMembershipsPage = domain.OrgMembershipsPage
+
```
duolok: here

auth/service.go
```diff
@@ -8,21 +8,27 @@ import (
...
+	Admin   = domain.OrgAdmin
+	Owner   = domain.OrgOwner
+	Editor  = domain.OrgEditor
+	Viewer  = domain.OrgViewer
+	RootSub = domain.RootSub
+	OrgSub  = domain.OrgSub
 )
 
+// AuthzReq is an alias for the shared domain type.
+type AuthzReq = domain.AuthzReq
```
duolok: Might be a minor thing, but you can put this inside a `type()` as well, so its consistent with the rest of the services.

things/profiles.go
```diff
@@ -5,37 +5,19 @@ package things
...
 
-// ProfilesPage contains page related metadata as well as list of profiles that
-// belong to this page.
-type ProfilesPage struct {
-	Total    uint64
-	Profiles []Profile
-}
+// Config and Transformer are aliases for the shared domain types.
+type Config = domain.Config
+type Transformer = domain.Transformer
```
duolok: Group aliases

auth/invites.go
```diff
@@ -17,14 +17,12 @@ import (
 var ErrInvalidInviteResponse = errors.New("invalid invite response action")
 var ErrGroupsDifferingOrgs = errors.New("groups belong to differing organizations")
```
majabirmancevic: not related to this, but you could put them under one var ()

pkg/domain/things.go
bool3max: Does it matter? Formatting tools (`gofmt`) leave it like this even if it's the only import, in fact I think most prefer it to always be a block, even if there's only one import, as adding more later on is easier

readers/service.go
```diff
@@ -91,10 +88,10 @@ func (rs *readersService) ListJSONMessages(ctx context.Context, token string, ke
 	return rs.json.Retrieve(ctx, rpm)
 }
 
-func (rs *readersService) ListSenMLMessages(ctx context.Context, token string, key things.ThingKey, rpm SenMLPageMetadata) (SenMLMessagesPage, error) {
+func (rs *readersService) ListSenMLMessages(ctx context.Context, token string, key domain.ThingKey, rpm SenMLPageMetadata) (SenMLMessagesPage, error) {
 	switch {
 	case rpm.Publisher != "":
-		_, err := rs.thingc.CanUserAccessThing(ctx, &protomfx.UserAccessReq{Token: token, Id: rpm.Publisher, Action: auth.Viewer})
+		_, err := rs.thingc.CanUserAccessThing(ctx, &protomfx.UserAccessReq{Token: token, Id: rpm.Publisher, Action: domain.OrgViewer})
```
majabirmancevic: fix here and in other methods

majabirmancevic (approved):

duolok (approved):
LGTM

manuio (approved):
LGTM

PR #1125, opened 2026-03-11
MF-1124 - Use dedicated error for missing thing key on update

### LINKED ISSUE #1124: Use dedicated error for missing thing key on update
Currently, updating a thing’s key returns a generic bearer/auth error, even though the key is supplied as part of the request body rather than as an auth credential. Introduce and use a dedicated `ErrMissingThingKey` for `updateThingReq` validation so that missing thing keys in update payloads are reported correctly and not conflated with authentication failures.

majabirmancevic:
Resolves #1124


anna02272 (approved):

duolok (approved):
LGTM

manuio (approved):
LGTM

Issue #1118, opened 2026-03-09
Implement machine‑to‑machine (M2M) communication


majabirmancevic:
Add first‑class support for machine‑to‑machine (M2M) communication, so one thing can send messages/commands directly to another thing in a controlled way.

majabirmancevic:
## Future consideration

**User-defined thing types and policies** — the current thing types (`controller`, `sensor`, 
`actuator`, etc.) and their command relationships are hardcoded platform defaults. This works 
well for the built-in types but limits flexibility for custom use cases (e.g. `camera`, 
`thermostat`, `plc`). 

Supporting user-defined types would require user-defined policies alongside them — defining 
which custom types can command which. This is a significant feature (type management + policy 
storage + runtime validation) and should be considered when a concrete use case drives it.

manuio:
I'd keep the thing `type` for RBAC and add a new `tags` field so users can define the type of the Thing. Eventually I'd even rename `type` to `role`.

PR #1132, opened 2026-03-12
MF-1118 - Implement machine‑to‑machine (M2M) authorization

majabirmancevic:
Implements machine‑to‑machine (M2M) authorization — things can send messages and commands to other things within the same group, subject to a capability policy based on their type.

**Authorization policy:**
- `command` — restricted by publisher type (controllers and gateways can command sensors, actuators and devices; sensors and actuators cannot issue commands)
- `message` — unrestricted within a group; subscribe-side enforcement in `AuthSubscribe` is sufficient, a publish check would be a redundant gRPC round-trip with no real security gain

**Changes:**
- Add static capability matrix in `things/policy.go` defining which thing types can command which other types
- Expose `CanThingCommand` gRPC method to evaluate M2M command authorization
- Enforce M2M policy in the MQTT adapter on publish to `things/{recipientID}/commands` and `groups/{groupID}/commands`
- Document thing types and the capability matrix in `things/README.md`

Related to #1118


things/service.go
```diff
@@ -628,6 +638,36 @@ func (ts *thingsService) CanThingAccessGroup(ctx context.Context, req ThingAcces
...
+	}
+
+	// Publisher and recipient must belong to the same group.
+	if publisher.GroupID != recipient.GroupID {
+		return errors.ErrAuthorization
+	}
+
+	switch req.Action {
+	case ActionCommand:
+		return canCommand(ctx, publisher.Type, recipient.Type)
```
majabirmancevic: Use type of thing from Redis cache: https://github.com/MainfluxLabs/mainflux/pull/1137

things/service.go
```diff
@@ -636,6 +644,28 @@ func (ts *thingsService) CanThingAccessGroup(ctx context.Context, req ThingAcces
 	return nil
 }
 
+func (ts *thingsService) CanThingCommand(ctx context.Context, req ThingCommandReq) error {
+	if req.PublisherID == "" || req.RecipientID == "" {
```
anna02272: Do we need this check? Will it ever trigger since `req.validate()` already ensures if IDs are not empty?

duolok (dismissed):

things/service.go
majabirmancevic: not really need

things/policy.go
```diff
@@ -0,0 +1,35 @@
...
+func CanCommand(publisherType, recipientType string) error {
+	allowed, ok := thingPolicies[publisherType]
+	if !ok || !slices.Contains(allowed, recipientType) {
+		return errors.ErrAuthorization
+	}
+	return nil
+}
+
+// CanGroupCommand checks if the publisher has any command authority,
+// with no recipient type since the command broadcasts to all things in a group.
```
majabirmancevic: > `CanCommand` validates the recipient's type — a controller can't accidentally command another controller.
> `CanGroupCommand` only validates the publisher — it doesn't check what types are actually in the group, so a controller broadcasting to a group would reach other controllers too, even though thingPolicies doesn't allow that pairing.
> 
> To enforce the same per-type check for group commands, we'd need to know the types of all things in the group at authorization time — a DB call.

Facts:
- Group membership already implies a trust boundary.
- Pre-validating every member's type before a broadcast is an O(n) auth check that scales poorly.
- Things that can't act on a command simply ignore it — this is idiomatic in IoT/messaging systems (consumers filter, not publishers).

So, keep it coarse, but document the intent explicitly - improve the comment.

anna02272 (approved):

duolok (approved):
LGTM

manuio (approved):
LGTM

PR #1137, opened 2026-03-13
MF-1135 - Add thing type to Redis cache

### LINKED ISSUE #1135: Add a type of thing to the Redis cache

majabirmancevic:
Resolves #1135

### DIFF
--- a/things/service.go
+++ b/things/service.go
@@ -261,6 +261,12 @@ func (ts *thingsService) UpdateThing(ctx context.Context, token string, thing Th
 		return err
 	}
 
+	if thing.Type != "" {
+		if err := ts.thingCache.RemoveType(ctx, thing.ID); err != nil {
+			return err
+		}
+	}
+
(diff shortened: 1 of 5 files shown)


things/service.go
```diff
@@ -261,6 +261,12 @@ func (ts *thingsService) UpdateThing(ctx context.Context, token string, thing Th
 		return err
 	}
 
+	if thing.Type != "" {
+		if err := ts.thingCache.RemoveType(ctx, thing.ID); err != nil {
+			return err
+		}
+	}
```
manuio: Is this case possible? Shouldn't we overwrite by cache key?

manuio (changes requested):

things/service.go
majabirmancevic: right, I forgot that the type is required

anna02272 (approved):

manuio (approved):
LGTM

PR #1154, opened 2026-03-18
MF - 1153 - Allow OAuth login for existing users and invite-based registration when self-registration is disabled

### LINKED ISSUE #1153: OAuth flow blocked entirely when self-registration is disabled
When `MF_USERS_SELF_REGISTER_ENABLED=false`, OAuth endpoints (`/users/oauth/:provider`) return `403 Forbidden` for all requests.

This causes two issues:

* Existing OAuth users cannot log in because the check happens before verifying if the user already exists.
* Invited users cannot register via OAuth, since the invite flow only supports email/password.

## Expected behavior

* Existing OAuth users should always be able to log in.
* Invited users should be able to complete registration via OAuth (using `invite_id`).
* Only new user creation without a valid invite should be blocked when self-registration is disabled.

anna02272:
Resolves #1153

### DIFF
--- a/users/service.go
+++ b/users/service.go
@@ -604,81 +604,79 @@ func (svc usersService) fetchGitHubUser(ctx context.Context, code, verifier stri
 	return email, providerUserID, nil
 }
 
+// handleIdentity resolves or creates a User for an OAuth login flow.
+// It first checks for an existing provider identity. If found, it returns the
+// linked user (syncing their e-mail if it changed). Otherwise, it looks up or
+// creates a user by e-mail, handling invite validation if needed, and saves a
+// new identity record linking the provider to that user.
 func (svc usersService) handleIdentity(ctx context.Context, provider, email, providerUserID, inviteID, redirectPath string) (User, error) {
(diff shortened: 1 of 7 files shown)


users/service.go
bool3max, line 684: This entire method is pretty unreadable. Instead of the long if-else blocks, you could return early in multiple places: 

e.g.:

```go
var user User

if identity.UserID != "" {
    // ...
    
    // Update e-mail
    if user.Email != email {
        // ...
    }

    return user, nil
}
```

and afterwards, in what was previously the big `else` block, you could simplify it to e.g.

```go
user, err = svc.users.RetrieveByEmail(ctx, email)
if err != nil && !errors.Contains(err, dbutil.ErrNotFound) {
    // return early if it's an error we can't handle
    return User{}, err
}

if errors.Contains(err, dbutil.ErrNotFound) {
    // ...
}
```

etc... Also, considering how much work the helper does, I think it'd be worth it to document the entire flow in a comment.

bool3max (changes requested):

bool3max (approved):

manuio (approved):
LGTM

PR #1167, opened 2026-03-20
MF-1159 - Fix converters: message timestamp

### LINKED ISSUE #1159: converters: message timestamp not being correctly read from "created" CSV field
Converters seem to currently treat the `"created"` CSV field specially, but it ends up as part of the message payload and not the message creation timestamp.

e.g. publishing the following CSV file:

```csv
created,data
12345,"data1"
```

Results in the following JSON message:

```json
{
    ...
    "created": <unix timestamp of message publish date>,
    "payload": {
        "created": 12345,
        "data": "data1"
    }
}
```

In other words the `created` CSV field isn't being correctly interpreted as the message timestamp.

### LINKED ISSUE #1166: Fix missing Content-Type header on error responses
apiutil.EncodeError does not send Content-Type: application/json

EncodeError calls w.WriteHeader() before WriteErrorResponse sets w.Header().Set("Content-Type", "application/json"). Since headers must be set before WriteHeader, the Content-Type header is silently dropped on all error responses across all services.

manuio:
Resolves #1159 
Resolves #1166

### DIFF
--- a/converters/service.go
+++ b/converters/service.go
@@ -93,6 +93,17 @@ func (as *adapterService) PublishSenMLMessages(ctx context.Context, key string, 
 }
 
 func (as *adapterService) PublishJSONMessages(ctx context.Context, key string, csvLines [][]string) error {
+	pcr := &protomfx.ThingKey{
+		Type:  things.KeyTypeInternal,
+		Value: key,
+	}
+	pc, err := as.things.GetPubConfigByKey(ctx, pcr)
+	if err != nil {
+		return err
+	}
+
+	timeField := pc.GetProfileConfig().GetTransformer().GetTimeField()
+
 	msg := protomfx.Message{
 		Protocol: protocol,
 		Created:  time.Now().UnixNano(),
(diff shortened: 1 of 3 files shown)


duolok (dismissed):
LGTM

converters/service.go
```diff
@@ -62,50 +62,64 @@ func (as *adapterService) PublishSenMLMessages(ctx context.Context, key string,
...
 	}
 	counter := 0
 	keys := csvLines[0][1:]
 	msgs := []map[string]any{}
-	createdIdx := slices.Index(csvLines[0], "created")
-	if createdIdx == -1 {
-		return ErrInvalidTimeField
-	}
-
+	createdIdx := slices.Index(csvLines[0], timeField)
```
bool3max: Maybe rename it to `timeFieldIdx`?

duolok (approved):
LGTM

majabirmancevic (approved):

PR #1170, opened 2026-03-20
MF-1140 - Automatic Key Rotation

### LINKED ISSUE #1140: Automatic Certificate Rotation for Things
The certs service already supports manual renewal and revocation of X.509 certificates. What's missing is automated rotation — a background scheduler that periodically queries certificates approaching expiration (within the existing 30-day renewal window), automatically triggers RenewCert, and notifies the Thing of its new credentials via the message bus. This ensures mTLS-connected Things never go offline due to expired certificates without requiring manual operator intervention.

duolok:
## Key changes:
  - Add CertScheduler that periodically checks for certificates expiring within 30 days, auto-renews them via PKI, and notifies Things of new credentials via NATS
  - Scheduler runs immediately on startup then repeats on a configurable cron schedule (MF_CERTS_CHECK_TIME)
  - Add GET /certs/:serial/download endpoint that returns full cert data and marks it as downloaded
  - Downloading is one-time only — subsequent attempts return 403

Resolves #1140


cmd/certs/main.go
```diff
@@ -60,6 +62,9 @@ const (
 	defThingsGRPCURL     = "localhost:8183"
 	defThingsGRPCTimeout = "1s"
 
+	defBrokerURL         = "nats://localhost:4222"
+	defCertCheckTime     = "02:00"
```
duolok: Daily time for certificate renewal. Each 24h it checks for outdated certificates at `defCertCheckTime`.

certs/api/logging.go
```diff
@@ -39,6 +39,19 @@ func (lm *loggingMiddleware) IssueCert(ctx context.Context, token, thingID, ttl
 	return lm.svc.IssueCert(ctx, token, thingID, ttl, keyBits, keyType)
 }
 
+func (lm *loggingMiddleware) RotateCert(ctx context.Context, token, serial, thingID, ttl string, keyBits int, keyType string) (_ certs.Cert, err error) {
+	defer func(begin time.Time) {
+		message := fmt.Sprintf("Method rotate_cert for serial %s took %s to complete", serial, time.Since(begin))
```
manuio: log thingID

certs/api/requests.go
```diff
@@ -31,6 +31,35 @@ func (req addCertsReq) validate() error {
...
+
+	if req.serial == "" {
+		return apiutil.ErrMissingSerial
+	}
+
+	if req.ThingID == "" {
+		return apiutil.ErrMissingThingID
+	}
+
+	if req.TTL == "" {
```
manuio, line 64: Check KeyBits and KeyType

certs/api/transport.go
```diff
@@ -35,6 +35,13 @@ func MakeHandler(svc certs.Service, tracer opentracing.Tracer, pkiAgent pki.Agen
 		opts...,
 	))
 
+	r.Post("/certs/rotate/:serial", kithttp.NewServer(
```
manuio: `certs/:serial/rotate"`

cmd/certs/main.go
```diff
@@ -86,6 +91,9 @@ const (
 	envAuthGRPCTimeout   = "MF_AUTH_GRPC_TIMEOUT"
 	envThingsGRPCURL     = "MF_THINGS_GRPC_URL"
 	envThingsGRPCTimeout = "MF_THINGS_GRPC_TIMEOUT"
+	envBrokerURL     = "MF_BROKER_URL"
+	envCertCheckTime = "MF_CERTS_CHECK_TIME"
```
manuio: Fix the indentation

manuio (changes requested):

api/openapi/certs.yml
```diff
@@ -197,6 +264,109 @@ components:
           description: Unique identifier of the thing the certificate was issued for.
         certificate:
           type: string
+<<<<<<< HEAD
```
manuio: Fix the merge

api/openapi/certs.yml
```diff
@@ -81,6 +100,54 @@ paths:
           description: Certificate not found.
         '500':
           $ref: "#/components/responses/ServiceError"
+<<<<<<< HEAD
```
manuio: fix the merge

api/openapi/certs.yml
```diff
@@ -254,6 +425,50 @@ components:
               - key_bits
               - key_type
             properties:
+<<<<<<< HEAD
```
manuio: and here

manuio (changes requested):

api/openapi/certs.yml
```diff
@@ -81,6 +81,51 @@ paths:
           description: Certificate not found.
         '500':
           $ref: "#/components/responses/ServiceError"
+  /certs/{certID}/rotate:
```
majabirmancevic: certId

api/openapi/certs.yml
```diff
@@ -133,6 +178,13 @@ components:
       schema:
         type: string
       required: true
+    CertID:
```
majabirmancevic: CertId

certs/service.go
```diff
@@ -148,6 +158,15 @@ func (cs *certsService) IssueCert(ctx context.Context, token, thingID string, tt
 	return c, nil
 }
 
+func (cs *certsService) RotateCert(ctx context.Context, token, serial, thingID, ttl string, keyBits int, keyType string) (Cert, error) {
```
majabirmancevic, line 169: do all authenticated users have rights to manage certs?

maybe we should check `CanUserAccessThing`, to avoid issuing certs to other groups' things?
```
func (cs *certsService) IssueCert(ctx context.Context, token, thingID string, ttl string, keyBits int, keyType string) (Cert, error) {
	_, err := cs.auth.Identify(ctx, &protomfx.Token{Value: token})
	if err != nil {
		return Cert{}, err
	}
...
```

duolok: Good remark, thank you

bool3max (dismissed):

certs/api/endpoint.go
```diff
@@ -33,6 +33,30 @@ func issueCertEndpoint(svc certs.Service) endpoint.Endpoint {
...
 }
 
+func rotateCertEndpoint(svc certs.Service) endpoint.Endpoint {
+	return func(ctx context.Context, request any) (any, error) {
+		req := request.(rotateCertsReq)
+		if err := req.validate(); err != nil {
+			return nil, err
+		}
+
+		res, err := svc.RotateCert(ctx, req.token, req.serial, req.ThingID, req.TTL, req.KeyBits, req.KeyType)
```
majabirmancevic, line 35: does it make sense to use a `Cert` struct as a param instead of separate fields?

certs/api/endpoint.go
```diff
@@ -33,6 +33,30 @@ func issueCertEndpoint(svc certs.Service) endpoint.Endpoint {
...
+		if err != nil {
+			return issueCertRes{}, err
+		}
+
+		return issueCertRes{
```
majabirmancevic: maybe we could use some `buildCertResponse` helper to avoid mapping fields in all endpoints?

certs/api/logging.go
```diff
@@ -39,6 +39,19 @@ func (lm *loggingMiddleware) IssueCert(ctx context.Context, token, thingID, ttl
 	return lm.svc.IssueCert(ctx, token, thingID, ttl, keyBits, keyType)
 }
 
+func (lm *loggingMiddleware) RotateCert(ctx context.Context, token, serial, thingID, ttl string, keyBits int, keyType string) (_ certs.Cert, err error) {
+	defer func(begin time.Time) {
+		message := fmt.Sprintf("Method rotate_cert for serial %s thing id %s took %s to complete", serial, thingID, time.Since(begin))
```
majabirmancevic: for serial %s `and` thing id

certs/api/logging.go
```diff
@@ -91,6 +104,19 @@ func (lm *loggingMiddleware) RevokeCert(ctx context.Context, token, serial strin
 	return lm.svc.RevokeCert(ctx, token, serial)
 }
 
+func (lm *loggingMiddleware) DownloadCert(ctx context.Context, token, serial string) (c certs.Cert, err error) {
+	defer func(begin time.Time) {
+		message := fmt.Sprintf("Method download_cert for serial %s thing id %s took %s to complete", serial, c.ThingID, time.Since(begin))
```
majabirmancevic: and here

certs/api/requests.go
```diff
@@ -86,3 +126,21 @@ func (req *revokeReq) validate() error {
...
+	switch keyType {
+	case pki.RSAKeyType:
+		if keyBits != pki.RSAKeyBits2048 && keyBits != pki.RSAKeyBits4096 {
+			return apiutil.ErrMissingCertData
+		}
+	case pki.ECDSAKeyType:
+		if keyBits != pki.ECDSAKeyBits224 && keyBits != pki.ECDSAKeyBits256 && keyBits != pki.ECDSAKeyBits384 && keyBits != pki.ECDSAKeyBits521 {
+			return apiutil.ErrMissingCertData
+		}
+	case "":
```
majabirmancevic: do we need this case?

certs/service.go
```diff
@@ -148,6 +163,15 @@ func (cs *certsService) IssueCert(ctx context.Context, token, thingID string, tt
...
+	_, err := cs.RevokeCert(ctx, token, serial)
+	if err != nil {
+		return Cert{}, err
+	}
+
+	return cs.IssueCert(ctx, token, thingID, ttl, keyBits, keyType)
```
majabirmancevic: we should avoid calling service methods directly, because if we call `RevokeCert` and `IssueCert`, that means:

```
cs.auth.Identify is called twice (once per method)
cs.things.CanUserAccessThing is called twice
```

We could use unexported helpers that skip auth, then have the public methods do auth once

certs/api/endpoint.go
```diff
@@ -142,6 +126,7 @@ func downloadCertEndpoint(svc certs.Service) endpoint.Endpoint {
 			Serial:         cert.Serial,
 			ThingID:        cert.ThingID,
 		}, nil
+
```
majabirmancevic: no need for a blank line

certs/api/endpoint.go
majabirmancevic, line 123: `build` helper could also be used here

certs/service.go
```diff
@@ -133,6 +133,10 @@ func (cs *certsService) IssueCert(ctx context.Context, token, thingID string, tt
 		return Cert{}, errors.ErrAuthorization
```
majabirmancevic: or `return Cert{}, err` ?

certs/service.go
```diff
@@ -164,37 +168,46 @@ func (cs *certsService) IssueCert(ctx context.Context, token, thingID string, tt
 }
 
 func (cs *certsService) RotateCert(ctx context.Context, token, serial, thingID, ttl string, keyBits int, keyType string) (Cert, error) {
-	_, err := cs.RevokeCert(ctx, token, serial)
+	_, err := cs.auth.Identify(ctx, &protomfx.Token{Value: token})
```
majabirmancevic: just to be sure, do we need both `auth.Identify` and `things.CanUserAccessThing`?

duolok: Yes

certs/api/endpoint.go
duolok: No need for helper if it's used on a single place

majabirmancevic: I thought the same helper that you already created, since there are the same fields

certs/service.go
majabirmancevic: @duolok things service returns that error

certs/service.go
majabirmancevic: why? is this consistent with the rest of the code or is there a specific case?

manuio: For what we need it if we only check for an error and pass the token to `CanUserAccessThing`?

certs/service.go
manuio: We could also wrap it

certs/api/endpoint.go
duolok: But they are not the same, I could extract the common fields through another function but then attach the rest of the fields which would be ugly IMHO. I used a helper where the same type was used

certs/api/endpoint.go
```diff
@@ -82,6 +90,7 @@ func viewCertEndpoint(svc certs.Service) endpoint.Endpoint {
 			Serial:      cert.Serial,
 			ExpiresAt:   cert.ExpiresAt,
 			ThingID:     cert.ThingID,
+			Downloaded:  cert.Downloaded,
```
anna02272, line 94: Could we also include a downloaded flag in the certificates list endpoint response? It would allow the frontend to properly reflect the certificate state instead of relying on  403 error after the download attempt.

certs/api/logging.go
```diff
@@ -41,6 +42,19 @@ func (lm *loggingMiddleware) IssueCert(ctx context.Context, token, thingID, ttl
 	return lm.svc.IssueCert(ctx, token, thingID, ttl, keyBits, keyType)
 }
 
+func (lm *loggingMiddleware) RotateCert(ctx context.Context, token, serial, thingID, ttl string, keyBits int, keyType string) (_ certs.Cert, err error) {
+	defer func(begin time.Time) {
+		message := fmt.Sprintf("Method rotate_cert for serial %s and thing id %s took %s to complete", serial, thingID, time.Since(begin))
```
majabirmancevic: add `email := pkgauth.EmailFromToken(token)`

certs/api/logging.go
```diff
@@ -97,6 +111,19 @@ func (lm *loggingMiddleware) RevokeCert(ctx context.Context, token, serial strin
 	return lm.svc.RevokeCert(ctx, token, serial)
 }
 
+func (lm *loggingMiddleware) DownloadCert(ctx context.Context, token string, thingKey domain.ThingKey, serial string) (c certs.Cert, err error) {
+	defer func(begin time.Time) {
+		message := fmt.Sprintf("Method download_cert for serial %s and thing id %s took %s to complete", serial, c.ThingID, time.Since(begin))
```
majabirmancevic, line 117: and here

certs/api/logging.go
```diff
@@ -113,7 +114,8 @@ func (lm *loggingMiddleware) RevokeCert(ctx context.Context, token, serial strin
 
 func (lm *loggingMiddleware) DownloadCert(ctx context.Context, token string, thingKey domain.ThingKey, serial string) (c certs.Cert, err error) {
 	defer func(begin time.Time) {
-		message := fmt.Sprintf("Method download_cert for serial %s and thing id %s took %s to complete", serial, c.ThingID, time.Since(begin))
+		email := pkgauth.EmailFromToken(token)
```
majabirmancevic: should we check if `token` is empty (if `thingKey` is passed)?

duolok: Is there an example of that?

majabirmancevic: I mean, if the device wants to download, `thingKey` will be used for auth and the `token` will be empty, so the log will be `by user __`, right?

duolok: Check the latest commit

certs/api/requests.go
```diff
@@ -70,6 +111,25 @@ func (req *viewReq) validate() error {
...
+	token    string
+	thingKey domain.ThingKey
+}
+
+func (req *downloadReq) validate() error {
+	if req.serial == "" {
+		return apiutil.ErrMissingSerial
+	}
+
+	err := apiutil.ValidateThingKey(req.thingKey)
```
majabirmancevic: if `thingKey` is invalid (non-empty) we return `apiutil.ErrMissingAuth`?

certs/api/logging.go
majabirmancevic: > Check the latest commit

I'm not sure if we use "by thing (key)" anywhere, but we'll also log the`thing id` anyway, so we could just:

```
u := ""
if token != "" {
    u = fmt.Sprintf("by user %s", pkgauth.EmailFromToken(token))
}
message := fmt.Sprintf("Method download_cert %s, serial %s and thing id %s took %s to complete", u, serial, c.ThingID, time.Since(begin))
```

api/openapi/certs.yml
```diff
@@ -81,6 +81,59 @@ paths:
           description: Certificate not found.
         '500':
           $ref: "#/components/responses/ServiceError"
+  /certs/{certId}/rotate:
```
majabirmancevic: certId -> serial ?
```
   r.Post("/certs/:serial/rotate", kithttp.NewServer(
		kitot.TraceServer(tracer, "rotate_cert")(rotateCertEndpoint(svc)),
		decodeRotateCert,
		encodeResponse,
		opts...,
	))

	r.Get("/certs/:serial/download", kithttp.NewServer(
		kitot.TraceServer(tracer, "download_cert")(downloadCertEndpoint(svc)),
		decodeDownloadCert,
		encodeResponse,
		opts...,
	))
```

api/openapi/certs.yml
```diff
@@ -81,6 +81,59 @@ paths:
...
+      requestBody:
+        $ref: "#/components/requestBodies/RotateCertReq"
+      responses:
+        '201':
+          $ref: "#/components/responses/IssueCertRes"
+        '400':
+          description: Failed due to malformed JSON.
+        '401':
+          description: Missing or invalid access token provided.
+        '404':
```
majabirmancevic, line 105: add `'403'`

certs/api/logging.go
```diff
@@ -97,6 +112,19 @@ func (lm *loggingMiddleware) RevokeCert(ctx context.Context, token, serial strin
 	return lm.svc.RevokeCert(ctx, token, serial)
 }
 
+func (lm *loggingMiddleware) DownloadCert(ctx context.Context, thingKey domain.ThingKey, serial string) (c certs.Cert, err error) {
+	defer func(begin time.Time) {
+		message := fmt.Sprintf("Method download_cert by thing key, serial %s and thing id %s took %s to complete", serial, c.ThingID, time.Since(begin))
```
majabirmancevic: remove "by thing key"

certs/api/requests.go
```diff
@@ -70,6 +111,23 @@ func (req *viewReq) validate() error {
...
+	serial   string
+	thingKey domain.ThingKey
+}
+
+func (req *downloadReq) validate() error {
+	if req.serial == "" {
+		return apiutil.ErrMissingSerial
+	}
+
+	if req.thingKey.Value == "" {
```
majabirmancevic: there is no need for this check if we call `return apiutil.ValidateThingKey(req.thingKey)`

certs/service.go
```diff
@@ -152,29 +171,59 @@ func (cs *certsService) IssueCert(ctx context.Context, token, thingID string, tt
 	return c, nil
 }
 
-func (cs *certsService) RevokeCert(ctx context.Context, token, serial string) (Revoke, error) {
-	var revoke Revoke
+func (cs *certsService) RotateCert(ctx context.Context, token, serial, thingID, ttl string, keyBits int, keyType string) (Cert, error) {
+	if _, err := cs.auth.Identify(ctx, token); err != nil {
```
majabirmancevic: is this check necessary?

certs/service.go
```diff
@@ -152,29 +171,59 @@ func (cs *certsService) IssueCert(ctx context.Context, token, thingID string, tt
...
+
+	if _, err := cs.revokeCert(ctx, serial); err != nil {
+		return Cert{}, err
+	}
+
+	return cs.issueCert(ctx, thingID, ttl, keyBits, keyType)
+}
+
+func (cs *certsService) RevokeCert(ctx context.Context, token, serial string) (Revoke, error) {
 	_, err := cs.auth.Identify(ctx, token)
```
majabirmancevic: and here

majabirmancevic: actually in all methods where authorization is required, because the token is already checked in `CanUserAccess..`

duolok: True, we could probably remove all `Identify` checks from the service.

certs/scheduler.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	payload, err := json.Marshal(event)
+	if err != nil {
+		return err
+	}
+
+	msg := protomfx.Message{
+		Publisher:   cert.ThingID,
+		Protocol:    "certs",
```
manuio, line 181: We can keep it for now but we should use a different structure without protiocol here

certs/scheduler.go
```diff
@@ -0,0 +1,189 @@
...
+		Payload:     payload,
+		ContentType: "application/json",
+		Created:     time.Now().UnixNano(),
+	}
+
+	subject := nats.GetThingCommandsSubject(cert.ThingID, certRotationTopic)
+	return cs.publisher.Publish(subject, msg)
```
manuio, line 188: We should have a way to know if the device received the notification or not

certs/postgres/certs.go
```diff
@@ -127,6 +127,33 @@ func (cr certsRepository) Remove(ctx context.Context, serial string) error {
 	return nil
 }
 
+func (cr certsRepository) RetrieveExpiring(ctx context.Context, expiresWithin time.Duration) ([]certs.Cert, error) {
+	q := `SELECT thing_id, serial, expires_at, client_cert, client_key, issuing_ca,
+	      ca_chain, private_key_type, key_bits, downloaded FROM certs
+	      WHERE expires_at <= :deadline ORDER BY expires_at`
```
manuio: This could be dangerous depending on the size of the list

manuio (approved):
LGTM

majabirmancevic (dismissed):

certs/scheduler.go
manuio: https://github.com/MainfluxLabs/mainflux/issues/1209

certs/scheduler.go
manuio: https://github.com/MainfluxLabs/mproxy/issues/4

certs/postgres/certs.go
manuio: This returns every cert that has ever expired, not just those expiring within the requested window.
Use `WHERE expires_at > NOW() AND expires_at <= :deadline ORDER BY expires_at`

manuio (changes requested):

manuio (approved):
LGTM

majabirmancevic (approved):

PR #1171, opened 2026-03-20
MF-1096 - Refactor gRPC client interfaces to use domain types

### LINKED ISSUE #1096: Refactor gRPC client interfaces
gRPC clients for each service in the codebase are currently declared as the protobuf-generated interface types. The effects of this are multifold:

* Every client method accepts protobuf-generated request types and returns protobuf-generated response types, unnecessarily leaking them into the rest of the code. Call sites of gRPC client methods have to manually encode requests to protobuf types and decodes responses from protobuf types

* It forces unnecessary, duplicated, manual conversion of protobuf types to internal gRPC request/response types in the definitions of gRPC client methods, despite `go-kit` already being designed to handle these conversions

In a nutshell client code is harder to write and less efficient than it should be.

bool3max:
gRPC client interfaces are now defined on domain types instead of on protobuf-generated types, which makes client call sites much cleaner, easier to write, and more ergonomic.

For example, what was previously:

```go
req := protomfx.AssignRoleReq{
	Id:   admin.ID,
	Role: domain.RoleRootAdmin,
}

if _, err := svc.auth.AssignRole(ctx, &req); err != nil {
	return err
}
```

now becomes:

```go
if err := svc.auth.AssignRole(ctx, admin.ID, domain.RoleRootAdmin); err != nil {
	return err
}
```

(note also the absence of the useless, protobuf-specific second return value)

**In other words callers no longer have to import or care about protobuf types at all**. This also has the added benefit of resolving the redundant back-and-forth type conversions in gRPC client implementations discussed in the [issue](https://github.com/MainfluxLabs/mainflux/issues/1096):

Before (`master`):

https://github.com/MainfluxLabs/mainflux/blob/6933a9489dc4daea78f9af78a0768863da3139c7/things/api/grpc/client.go#L273-L284

... and after:

```go
func (client grpcClient) Identify(ctx context.Context, key domain.ThingKey) (string, error) {
	ctx, cancel := context.WithTimeout(ctx, client.timeout)
	defer cancel()

	res, err := client.identify(ctx, thingKey{value: key.Value, keyType: key.Type})
	if err != nil {
		return "", err
	}

	ir := res.(identityRes)

	return ir.id, nil
}
```

### Profile Config domain type changes

Domain types which embedded a Profile config type (namely `Profile` itself but also `PubConfigInfo`) previously did so using a `map[string]any` type, which made accesses of profile config properties awkward. This has been changed so that they now correctly use the proper `*ProfileConfig` type all throughout.

Depends on PR #1114. Closes #1096.


converters/service.go
```diff
@@ -93,16 +94,17 @@ func (as *adapterService) PublishSenMLMessages(ctx context.Context, key string,
...
 	}
-	pc, err := as.things.GetPubConfigByKey(ctx, pcr)
+
+	pc, err := as.things.GetPubConfigByKey(ctx, thKey)
 	if err != nil {
 		return err
 	}
 
-	timeField := pc.GetProfileConfig().GetTransformer().GetTimeField()
+	timeField := pc.ProfileConfig["transformer"].(map[string]any)["time_field"].(string)
```
bool3max: Previously the things service gRPC methods `GetPubConfigByKey` and `GetConfigByThing` were returning proto-generated `Config` and `Transformer` structured types. 

Now they return the `PubConfigInfo` domain type which for some reason holds the Config as a `map[string]any` (which is not as easy to access), even though the domain defines proper `Config` and `Transformer` types?

We should probably modify `PubConfigInfo` to hold these concrete types.

majabirmancevic: yeah, and probably here as well
https://github.com/MainfluxLabs/mainflux/blob/a0a5338a336f179b1e5d5be0caa19563a54137e4/things/profiles.go#L16

manuio (dismissed):
LGTM

http/adapter.go
```diff
@@ -89,25 +86,25 @@ func (as *adapterService) SendCommandToGroup(ctx context.Context, token, groupID
 }
 
 func (as *adapterService) SendCommandToThingByKey(ctx context.Context, key domain.ThingKey, thingID string, message protomfx.Message) error {
-	res, err := as.things.Identify(ctx, &protomfx.ThingKey{Value: key.Value, Type: key.Type})
+	publishedID, err := as.things.Identify(ctx, key)
```
majabirmancevic: publisherID

things/postgres/profiles.go
```diff
@@ -323,23 +334,73 @@ type dbProfile struct {
 }
 
 func toDBProfile(pr things.Profile) dbProfile {
+	var config dbJSONB
+
+	if pr.Config != nil {
+		config = map[string]any{
+			"content_type": pr.Config.ContentType,
+			"transformer": map[string]any{
```
majabirmancevic, line 342: `Transformer` has JSON struct tags so `pr.Config.Transformer` is enough?

bool3max: `dbProfile` stores the config as a `dBJSONB` type which is a `map[string]any`

pkg/domain/users.go
```diff
@@ -47,6 +48,19 @@ type UsersPage struct {
 	Users []User `json:"users"`
 }
 
+// PageMetadata contains page metadata that helps navigation.
+type UsersPageMetadata struct {
```
majabirmancevic, line 52: all PageMetadata structs are defined within each service, so this introduces inconsistency?

certs/service.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) Mainflux
+// copyright (c) mainflux
```
duolok: Probably didn't mean to change this

pkg/domain/users.go
bool3max: The users service is currently the only one which exposes gRPC methods that accept a `PageMetadata` as an argument, and so to avoid cross-service imports (between `users` and services calling into it), the type has to live in a separate shared package.

pkg/domain/things.go
```diff
@@ -52,7 +53,7 @@ type Profile struct {
 	ID       string         `json:"id,omitempty"`
 	GroupID  string         `json:"group_id,omitempty"`
 	Name     string         `json:"name,omitempty"`
-	Config   map[string]any `json:"config,omitempty"`
+	Config   *ProfileConfig `json:"config,omitempty"`
```
majabirmancevic, line 56: do we need a pointer here?

ws/adapter.go
```diff
@@ -67,7 +68,7 @@ func (svc *adapterService) Publish(ctx context.Context, key domain.ThingKey, msg
 		return messaging.ErrPublishMessage
 	}
 
-	if err := messaging.FormatMessage(pc, &msg); err != nil {
+	if err := messaging.FormatMessage(protoutil.PubConfigInfoToProto(pc), &msg); err != nil {
```
majabirmancevic: maybe in the next PR we can refactor this helper to pass `domain.PubConfigInfo` as a param instead of converting `protoutil.PubConfigInfoToProto(pc)`

bool3max: Good idea. I just changed it, and the `pkg/transformers/json.TransformPayload` method to take domain types instead. Some call sites in downlinks and modbus needed some minor changes, but it should be all good.

pkg/domain/things.go
bool3max: A profile doesn't necessarily have a config attached to it. In those cases, the `Config` field here is `nil`.

majabirmancevic (approved):

manuio (approved):
LGTM

PR #1176, opened 2026-03-24
MF-1139 - Add CRL to Certs

### LINKED ISSUE #1139: Add CRL to Certs

duolok:
This PR adds CRL support to the platform. It could be tested with the following request after issuing a cert:

```bash
curl -k -v -X POST "https://localhost/http/messages" --cert test-device.crt --key test-device.key -H "Content-Type: application/json" -d '{"n":"humidity","v":1234}'
```

The requests only sends the message using mTLS. After that revoke the cert and try sending the request again. It should fail as revoked cert has been added to the CRL.

Resolves #1139


docker/addons/certs/docker-compose.yml
```diff
@@ -57,15 +58,17 @@ services:
       MF_CERTS_HTTP_PORT: ${MF_CERTS_HTTP_PORT}
       MF_CERTS_SERVER_CERT: ${MF_CERTS_SERVER_CERT}
       MF_CERTS_SERVER_KEY: ${MF_CERTS_SERVER_KEY}
-      MF_CERTS_SIGN_CA_PATH: ${MF_CERTS_SIGN_CA_PATH}
-      MF_CERTS_SIGN_CA_KEY_PATH: ${MF_CERTS_SIGN_CA_KEY_PATH}
+      MF_CERTS_SIGN_CA_PATH: /etc/ssl/certs/ca.crt
+      MF_CERTS_SIGN_CA_KEY_PATH: /etc/ssl/certs/ca.key
```
manuio: Are we using this ENVARS anymore?

certs/service.go
```diff
@@ -22,6 +23,9 @@ var (
 	ErrFailedCertRevocation = errors.New("failed to revoke certificate")
 
 	errFailedToRemoveCertFromDB = errors.New("failed to remove cert serial from db")
+
+	// ErrFailedCRLGeneration indicates failure to generate CRL file.
+	ErrFailedCRLGeneration = errors.New("failed to generate CRL")
```
manuio: Do we need to export the error? `Generate` or `Regenerate`?

docker/ssl/certs/ca.crt
```diff
@@ -1,22 +1,31 @@
 -----BEGIN CERTIFICATE-----
-MIIDjzCCAnegAwIBAgIUQ1AagVQXCuOIzmGXm+KhsbyBc18wDQYJKoZIhvcNAQEN
```
manuio, line 2: Here and for all other certs, do we need new ones?

manuio (changes requested):

docker/ssl/certs/ca.crt
duolok: I've made some changes in certs `Makefile`, so It's important that we update the certs in order to have functional mTLS locally. I don't see why not, if they are not updated you'll have to run `make all` anyway.

manuio: Because this certs are an example, they shoul de regenerated in any case, right?

duolok: You'd still want to have an example that works, rather than an outdated one.

docker/docker-compose.yml
```diff
@@ -848,18 +850,20 @@ services:
       MF_CERTS_HTTP_PORT: ${MF_CERTS_HTTP_PORT}
       MF_CERTS_SERVER_CERT: ${MF_CERTS_SERVER_CERT}
       MF_CERTS_SERVER_KEY: ${MF_CERTS_SERVER_KEY}
-      MF_CERTS_SIGN_CA_PATH: ${MF_CERTS_SIGN_CA_PATH}
-      MF_CERTS_SIGN_CA_KEY_PATH: ${MF_CERTS_SIGN_CA_KEY_PATH}
+      MF_CERTS_SIGN_CA_PATH: /etc/ssl/certs/ca.crt
```
manuio: Why don't we use ENVARS anymore?

manuio (changes requested):

docker/.env
```diff
@@ -1,7 +1,4 @@
 # Docker: Environment variables in Compose
- 
-# Auth mode: "key" for API key auth, "x509" for mTLS client certificate auth
-AUTH=x509
```
manuio: Shouldn't we keep the ENVAR?

manuio (changes requested):

manuio (dismissed):
LGTM

majabirmancevic (approved):

manuio (approved):
LGTM

PR #1178, opened 2026-03-24
MF-1168 - Publish MQTT commands to corresponding NATS subject

### LINKED ISSUE #1168: Fix MQTT commands being forwarded as messages

majabirmancevic:
Resolves #1168

### DIFF
--- a/mqtt/handler.go
+++ b/mqtt/handler.go
@@ -261,12 +261,12 @@ func parseTopic(topic, publisherID string) (subject, subtopic string, err error)
 			}
 		}
 		switch {
-		case suffix == topicSuffixCommands && prefix == topicPrefixThings:
+		case prefix == topicPrefixThings && suffix == topicSuffixCommands:
 			return nats.GetThingCommandsSubject(id, rest), rest, nil
-		case suffix == topicSuffixCommands && prefix == topicPrefixGroups:
+		case prefix == topicPrefixGroups && suffix == topicSuffixCommands:
 			return nats.GetGroupCommandsSubject(id, rest), rest, nil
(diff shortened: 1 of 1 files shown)


mqtt/handler.go
```diff
@@ -241,11 +241,43 @@ func (h *handler) Publish(c *session.Client, topic *string, payload *[]byte) {
...
+	if len(parts) >= 3 && parts[1] != "" {
+		prefix, id, suffix := parts[0], parts[1], parts[2]
+		rest := ""
+		if len(parts) > 3 {
+			if rest, err = messaging.NormalizeSubtopic(strings.Join(parts[3:], "/")); err != nil {
+				return "", "", err
+			}
+		}
+		switch {
+		case suffix == topicSuffixCommands && prefix == topicPrefixThings:
```
manuio: Can we check if `prefix != topicPrefixThings` and return before the switch case?

manuio (changes requested):

mqtt/handler.go
majabirmancevic: If you mean both prefixes, it works, but it's purely cosmetic; the prefix check is already implicit in every switch case — adding an explicit guard before it is redundant indirection. The current flat structure is easier to scan, in my opinion. 

- One thing worth noting: the `messages` case can't be dropped to default even though it looks like it could — it routes to the topic's `target thing` subject using **id from the topic**, while the default uses `publisherID`, so they produce different NATS subjects.

manuio (approved):
LGTM

anna02272 (approved):

duolok (approved):

PR #1180, opened 2026-03-25
MF-1177 - Fix changing scheduler frequency has no effect

### LINKED ISSUE #1177: Changing scheduler frequency has no effect
Updating a modbus client's scheduler frequency leaves the old polling task still running. RemoveCronEntry looks up the cron instance by the timezone passed by the caller, but if it doesn't match the key used at scheduling time, c.Remove is silently skipped and the old task keeps firing.

manuio:
Resolves #1177


bool3max (approved):

duolok (approved):
LGTM

PR #1184, opened 2026-03-27
MF-1179 - Polish logs and errors in MQTT handler 

### LINKED ISSUE #1179: Polish logs in MQTT handler
`LogErr*` string constants and `errors.Wrap` are used interchangeably for the same purpose. 
Remove the constants and unify all error logging to use `errors.Wrap`.

majabirmancevic:
Resolves #1179


mqtt/handler_test.go
```diff
@@ -293,26 +293,19 @@ func TestPublish(t *testing.T) {
 			payload: payload,
 			logMsg:  mqtt.ErrClientNotInitialized.Error(),
 		},
-		{
```
majabirmancevic, line 296: The only invalid topics are the ones that fail `NormalizeSubtopic` (those are already covered by the `malformedSubtopics (%)` and `wrongCharSubtopics (>)` cases.)

manuio (approved):
LGTM

Issue #991, opened 2025-12-24
Send a command to a device and a device group


majabirmancevic:
Implement the functionality of sending commands from the UI directly to a specific device or group of devices.
Using adapters, send commands to NATS where the `mqtt-adapter` will consume the commands and forward them to the mqtt broker using the `forwarder`. 
Devices subscribed to certain topics consume those commands and perform the necessary actions.

PR #1186, opened 2026-03-31
MF-991 - Send commands via ws adapter

majabirmancevic:
Related to #991


majabirmancevic:
add docs

ws/api/endpoints.go
```diff
@@ -43,51 +44,122 @@ func handshake(svc ws.Service) http.HandlerFunc {
...
+}
+
+func decodeCommandRequest(r *http.Request) (cmdConnReq, error) {
+	id := bone.GetValue(r, apiutil.IDKey)
+	if id == "" {
+		return cmdConnReq{}, errors.ErrMalformedEntity
+	}
+
+	var subtopicPath string
+	if idx := strings.LastIndex(r.URL.Path, "/commands/"); idx >= 0 {
```
manuio: Use  `strings.Index`  instead of `strings.LastIndex` so we can eventually support `/things/id/commands/pump/commands/start`

manuio (changes requested):

ws/api/endpoints.go
```diff
@@ -43,51 +44,122 @@ func handshake(svc ws.Service) http.HandlerFunc {
...
 		return getConnByKey{}, err
 	}
-
-	req := getConnByKey{
-		ThingKey: thingKey,
+	if tk.Value == "" {
+		return getConnByKey{}, apiutil.ErrMissingAuth
 	}
 
 	subtopic, err := messaging.NormalizeSubtopic(r.RequestURI)
```
majabirmancevic: use the extracted subtopic instead of `r.RequestURI` here?

ws/api/endpoints.go
```diff
@@ -43,51 +44,122 @@ func handshake(svc ws.Service) http.HandlerFunc {
...
+func decodeMessagesRequest(r *http.Request) (getConnByKey, error) {
+	tk, err := extractThingKey(r)
+	if err != nil {
 		return getConnByKey{}, err
 	}
-
-	req := getConnByKey{
-		ThingKey: thingKey,
+	if tk.Value == "" {
+		return getConnByKey{}, apiutil.ErrMissingAuth
```
majabirmancevic: ErrMissingThingKey

manuio (dismissed):
LGTM

api/asyncapi/websocket.yml
```diff
@@ -0,0 +1,178 @@
...
+      - thingKey: []
+      - bearerToken: []
+
+channels:
+  /messages:
+    publish:
+      summary: Publish a message
+      description: Sends a message without a subtopic.
+      message:
+        $ref: '#/components/messages/SenMLMsg'
```
duolok: SenMLMessage

duolok: Replace throughout the entire file

api/asyncapi/websocket.yml
```diff
@@ -0,0 +1,178 @@
...
+      subtopic:
+        $ref: '#/components/parameters/subtopic'
+    publish:
+      summary: Send a command to a thing on a subtopic
+      description: |
+        Each message sent over the socket is published as a command to the specified thing on a subtopic.
+        Supports nested subtopics with `/`.
+        Authorized by thing key (M2M) or user bearer token.
+      message:
+        $ref: '#/components/messages/CommandMsg'
```
duolok: This should be CommandMessage then

ws/api/endpoints.go
```diff
@@ -19,9 +20,9 @@ import (
 	"github.com/gorilla/websocket"
 )
 
-func handshake(svc ws.Service) http.HandlerFunc {
+func messagesHandshake(svc ws.Service) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		req, err := decodeRequest(r)
+		req, err := decodeMessageRequest(r)
```
duolok, line 19: Not sure if this is a requirement of WebSockets, but should we have decoder functions inside `transport.go`?

ws/api/endpoints.go
```diff
@@ -43,51 +44,119 @@ func handshake(svc ws.Service) http.HandlerFunc {
...
+func groupCommandsHandshake(svc ws.Service) http.HandlerFunc {
+	return commandsHandshake(svc, processGroupCommands)
+}
 
-		thingKey = domain.ThingKey{
-			Value: queryKey[0],
-			Type:  queryKeyType[0],
+func commandsHandshake(svc ws.Service, processFn func(ws.Service, cmdConnReq, <-chan []byte)) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		req, err := decodeCommandRequest(r)
```
duolok, line 56: and here

ws/api/endpoints.go
```diff
@@ -43,51 +44,119 @@ func handshake(svc ws.Service) http.HandlerFunc {
...
+		req.token = token
+		return req, nil
+	}
+
+	return cmdConnReq{}, apiutil.ErrMissingAuth
+}
+
+// extractThingKey retrieves a ThingKey from the Authorization header,
+// falling back to the key and keyType query parameters.
+func extractThingKey(r *http.Request) domain.ThingKey {
```
duolok: then `extractThingKey` could also be moved to `transport.go`, or even better to `apiutil`

ws/api/endpoints.go
```diff
@@ -43,51 +44,119 @@ func handshake(svc ws.Service) http.HandlerFunc {
...
+	if len(queryKey) > 0 && len(queryKeyType) > 0 {
+		return domain.ThingKey{Value: queryKey[0], Type: queryKeyType[0]}
+	}
+
+	return domain.ThingKey{}
+}
+
+// extractBearerToken retrieves a bearer token from the Authorization header,
+// falling back to the token query parameter for browser WS clients that cannot set headers.
+func extractBearerToken(r *http.Request) string {
```
duolok: Could this also be moved to `apiutil`?

ws/api/endpoints.go
majabirmancevic: `decodeRequest` was done there, so I just followed that, but yes, we could have moved in `transport.go`

ws/api/endpoints.go
majabirmancevic: these `extractBearerToken` and `extractThingKey` are specific to the ws adapter because of the way the token/key is extracted, so we should keep them there

api/asyncapi/websocket.yml
```diff
@@ -0,0 +1,178 @@
...
+        format: uuid
+
+  securitySchemes:
+    thingKey:
+      type: httpApiKey
+      name: Authorization
+      in: header
+      description: |
+        Thing key authentication.
+        Header value: "Thing <thing_key>" or use query params: ?key=<key>&keyType=<type>
```
manuio, line 166: Why are we accepting this?

api/asyncapi/websocket.yml
```diff
@@ -0,0 +1,178 @@
...
+    bearerToken:
+      type: http
+      scheme: bearer
+      description: |
+        User bearer token authentication.
+        Header value: "Bearer <user_token>" or use query param: ?token=<user_token>
```
manuio, line 172: And this?

manuio (changes requested):

api/asyncapi/websocket.yml
majabirmancevic: that's how it was supported before (header and query param options), I guess that's because WebSocket connections from browsers can't always set headers, so query param auth is a common workaround, although it's not ideal..

duolok (approved):
LGTM

manuio (approved):
LGTM

PR #1189, opened 2026-04-01
MF-1175 - rules: scripts: add readers Lua API

### LINKED ISSUE #1175: rules: extend Lua scripting API with ability to fetch past messages from readers
... will require readers to expose a gRPC interface, which it currently doesn't.

bool3max:
* #### Exposed a gRPC server in the `readers` service
    * Endpoints are provided for the `ListJSONMessages` and `ListSenMLMessages` service methods only
* #### Added readers support to the Lua scripting API
    * ```lua
      mfx.list_messages(<message_kind>, <thing_key>, [<page_metadata>])
      ```
        * `<message_kind>`: one of `"json"` or `"senml"` for listing JSON and SenML messages, respectively
        * `<thing_key>`: a Lua table of the following structure: `{ type = "external"|"internal", value = "<key>"}`, _or_ `nil`, in which case messages from the publisher associated with the currently-executing script are fetched 
        * `[<page_metadata>]`: an optional table representing pagination metadata and 
filters (see the definitions of `domain.SenMLPageMetadata` and `domain.JSONPageMetadata`)
        * Returns:
            * `(table)` an array-like table containing all returned messages, or `nil` if error
            * `(number)` the _total_ number of messages available on the device (-1 if error)
            * _OPTIONAL_ `(string)` an error message in case of an error
    * Example - fetch the last 5 messages from specific sensor, compute an average of the values, and send an alarm if it exceeds a threshold:
    
         ```lua
         local temp_sensor_key = { type = "external", value = "02a7b477-dc14-430a-9e36-6a344f4a10a1" }

         messages, total, err = mfx.list_messages("senml", temp_sensor_key, { limit = 5 })

         if not messages then
             error(err)
         end

         local temp_sum = 0

         for idx, msg in ipairs(messages) do
             temp_sum = temp_sum + msg.value
         end

         local avg = temp_sum / #messages
         if avg > 32.5 then
             mfx.log("average temp " .. avg .. " exceeds threshold")
             mfx.create_alarm()
         end

###### TODO:

* [x] Implicit self-authentication with `nil` thing key - e.g. `mfx.list_messages("json", nil)` to fetch messages of Thing associated with the currently running script

Depends on PR #1171. Resolves #1175.


rules/lua_api.go
```diff
@@ -114,3 +123,378 @@ var luaLog = luaAPIFunc{
...
+					ls.PushInteger(0)
+					ls.PushString(err.Error())
+
+					return 3
+				}
+
+				// Return ({messages}, total)
+				ls.NewTable()
+				for idx, msg := range page.Messages {
+					pushSenMLMessageToLua(ls, msg.(senml.Message))
```
bool3max, line 212: The type assertion is necessary because `readers` represents returned messages (both JSON and SenML ones) with a `type Message any` type, which I think we should seriously rethink.

rules/lua_api.go
duolok: What does `ap` stand for?

bool3max: Great question, I've got no idea, I probably cut it off accidentally. It's likely supposed to end in "... publish it to NATS SMTP subject" or something similar

pkg/domain/readers.go
```diff
@@ -62,3 +64,9 @@ type JSONPageMetadata struct {
 	AggFields   []string `json:"agg_fields,omitempty"`
 	Dir         string   `json:"dir,omitempty"`
 }
+
+// ReadersClient specifies the API for querying messages from the readers service via gRPC.
+type ReadersClient interface {
```
manuio, line 69: Why is this in pkg repo?

manuio (changes requested):

pkg/domain/readers.go
bool3max: Because `readers` now exposes a gRPC interface, and other services (`rules` currently) has to import the client to interact with it. It's for the same reason why e.g. `AuthClient` or `ThingsClient` are also in `pkg/domain`.

manuio: ok!

manuio (changes requested):
LGTM

manuio (approved):
LGTM

duolok (approved):
LGTM

PR #1194, opened 2026-04-06
NOISSUE - auth: use proper protobuf type in gRPC client Issue method

bool3max:
For a pretty long time the `Issue` method of the `auth` service gRPC client seemed to decode the server response into the `UserIdentity` protobuf type, while the **server responds with the `Token` protobuf type**. 

This worked purely by coincidence, since the `Issue` client method only cared about the _first field of the response type_ (`id` in `UserIdentity`), which shares the field number and type with the first (and only) field of the actual response type returned by the server (`value` in `Token`).

This PR fixes the client side to decode the response into the proper `Token` protobuf type and then later into the `issueRes` gRPC response type.


majabirmancevic (approved):

manuio (approved):
LGTM

PR #1195, opened 2026-04-07
MF-1150 - Add level, status, and rule conditions to alarms

### LINKED ISSUE #1150: Add level, status, and rule conditions to alarms
Alarms currently have no structured classification or lifecycle state, making it impossible to filter or prioritize them at the API level.

Add three fields to the Alarm struct:

* condition — structure describing the condition (name, field, threshold,, comparator)
* Level int — severity set by the triggering rule or script (1=info, 2=warning, 3=minor, 4=major, 5=critical)
* Status string — lifecycle state (active, noted, cleared), defaults to active, transitioned via API

Requires DB migration and OpenAPI spec update.

majabirmancevic:
Resolves #1150 and #1200


consumers/alarms/service.go
```diff
@@ -195,6 +212,27 @@ func (as *alarmService) createAlarm(ctx context.Context, alarm *Alarm) error {
 
 }
 
+func extractRule(payload map[string]any) *RuleInfo {
+	ruleRaw, ok := payload["rule"]
+	if !ok {
+		return nil
+	}
+
+	delete(payload, "rule")
```
manuio: Why do we have to delete this?

consumers/alarms/service.go
```diff
@@ -195,6 +212,27 @@ func (as *alarmService) createAlarm(ctx context.Context, alarm *Alarm) error {
...
+	if err != nil {
+		return nil
+	}
+
+	var rule RuleInfo
+	if err := json.Unmarshal(b, &rule); err != nil {
+		return nil
+	}
+
+	return &rule
```
manuio: `return rule`

rules/api/http/rules/endpoint_test.go
```diff
@@ -352,6 +352,23 @@ func TestCreateRules(t *testing.T) {
 			status: http.StatusBadRequest,
 			size:   0,
 		},
+		{
+			desc:        "create rule with invalid alarm level",
+			auth:        token,
+			groupID:     groupID,
+			contentType: contentType,
+			body: toJSON(map[string]any{
```
manuio: Should we use a structure instead?

rules/api/http/rules/requests.go
```diff
@@ -86,6 +86,9 @@ func (req rule) validate() error {
 				return apiutil.ErrMissingActionID
 			}
 		case rules.ActionTypeAlarm:
+			if action.Level < 1 || action.Level > 5 {
```
manuio: Let's create consts for levels. BTW, we can cnsider having 100. Otherwise, we could use names for levels.

rules/lua_api.go
```diff
@@ -51,11 +51,18 @@ var luaSMTPNotify = luaAPIFunc{
...
 // Lua signature:
-// mfx.create_alarm()
+// mfx.create_alarm(level) where level is an integer (1=info, 2=warning, 3=minor, 4=major, 5=critical)
 // On success it returns true, nil. On failure, it returns false, <error_message>
 var luaAlarmCreate = luaAPIFunc{
 	fun: func(env *luaEnv) lua.Function {
 		return func(ls *lua.State) int {
+			level, ok := ls.ToInteger(1)
+			if !ok {
+				ls.PushBoolean(false)
```
manuio, line 66: What is this for?

rules/rules.go
```diff
@@ -22,15 +22,12 @@ type Rule struct {
...
-	Threshold  *float64 `json:"threshold"`
-}
+type Condition = domain.Condition
 
 type Action struct {
-	ID   string `json:"id"`
-	Type string `json:"type"`
+	ID    string `json:"id"`
+	Type  string `json:"type"`
+	Level int    `json:"level,omitempty"`
```
manuio: Why omitempty only here?

manuio (changes requested):

consumers/alarms/service.go
majabirmancevic: we extract it from the payload and map it to the structure - `alarm.Rule`, we store it as a separate field in the db, so we don't need it in the payload, because it would be redundant

rules/rules.go
majabirmancevic: `Level` is only meaningful for `ActionTypeAlarm`, for SMTP/SMPP actions it should be zero?

rules/lua_api.go
majabirmancevic: that's  level check, we need to pass level in subject for script-based alarms too?

manuio: Yes, why not? But what I meant is why we do `ls.PushBoolean(false)`?

rules/rules.go
manuio: Why not to include level of the alarm in the notification?

majabirmancevic: the initial idea (issue) was to add a few fields to Alarm, so we have a kind of classification, and on the other hand, alarm and notification are 2 types of actions, so "_to include level of the alarm in the notification_" doesn't make sense, right?

we could add (#later) a `level` to the notification/email body or subject (or somewhere for smpp), so in that case, I can remove "omitempty" and adjust request validation

rules/lua_api.go
majabirmancevic: aha, so when the `level` argument is missing or invalid, we return false and an error message to let the Lua script know the call failed (that's how it's done in the rest of the function)

consumers/alarms/service.go
majabirmancevic: `extractRule` returns `*RuleInfo`

- **update**: `extractRule` removed

duolok (dismissed):
LGTM

consumers/alarms/service.go
manuio: Let's find a better way to transferm alarm info in a separate PR

rules/rules.go
manuio: Right, there are different types of actions but they are similar. we can see notification as alarms that you send as ean email instead of storing it in the system. We can do it later but we can consider having a level of notification also.

consumers/alarms/api/http/converters.go
```diff
@@ -26,7 +26,7 @@ func ConvertToJSONFile(page alarms.AlarmsPage, timeFormat string) ([]byte, error
 			"script_id": a.ScriptID,
 			"subtopic":  a.Subtopic,
 			"protocol":  a.Protocol,
-			"payload":   a.Payload,
+			"rule":      a.Rule,
```
anna02272, line 29: We should add level and status field in export as well, right?

majabirmancevic: right!

consumers/alarms/postgres/alarms.go
```diff
@@ -118,23 +132,16 @@ func (ar *alarmRepository) RetrieveByGroup(ctx context.Context, groupID string,
 	dq := dbutil.GetDirQuery(pm.Dir)
 	olq := dbutil.GetOffsetLimitQuery(pm.Limit)
 
-	p, pq, err := dbutil.GetPayloadQuery(pm.Payload)
-	if err != nil {
-		return alarms.AlarmsPage{}, errors.Wrap(dbutil.ErrRetrieveEntity, err)
-	}
+	whereClause := dbutil.BuildWhereClause("group_id = :group_id")
```
anna02272, line 135: Could we extend retrieving where clause and add level so we can have filtering?
Example of what we need:

Image

Image

majabirmancevic: sure, I was thinking about that when I added order options, but honestly, I'd leave filtering for the next PR to reduce code changes, but we could open an issue for that

majabirmancevic: here it is https://github.com/MainfluxLabs/mainflux/issues/1205

duolok (dismissed):
LGTM

rules/README.md
```diff
@@ -42,12 +42,13 @@ Each condition compares a named field in the message payload against a numeric t
 
 Each action specifies what to do when a rule fires.
 
-| Field  | Description                                                                         |
-|--------|-------------------------------------------------------------------------------------|
-| `type` | Action type: `alarm`, `smtp`, or `smpp`                                             |
-| `id`   | Required for `smtp` and `smpp` types — the ID of the configured notifier to trigger |
+| Field   | Description                                                                                              |
+|---------|----------------------------------------------------------------------------------------------------------|
+| `type`  | Action type: `alarm`, `smtp`, or `smpp`                                                                  |
```
manuio: Align the indentation

rules/README.md
```diff
@@ -83,7 +84,7 @@ Each script execution receives an isolated Lua environment. The following global
 | Function                       | Returns             | Description                                                                    |
 |--------------------------------|---------------------|--------------------------------------------------------------------------------|
 | `mfx.smtp_notify(notifier_id)` | `bool[, error_msg]` | Triggers an SMTP notification via the specified notifier. Max 2 calls per run. |
-| `mfx.create_alarm()`           | `bool[, error_msg]` | Creates an alarm event attributed to this script. Max 1 call per run.          |
+| `mfx.create_alarm(level)`      | `bool[, error_msg]` | Creates an alarm event attributed to this script. Level is an integer (1=info, 2=warning, 3=minor, 4=major, 5=critical). Max 1 call per run. |
```
manuio: and here

manuio (changes requested):

consumers/alarms/postgres/init.go
```diff
@@ -72,6 +72,21 @@ func migrateDB(db *sqlx.DB) error {
 					`ALTER TABLE alarms DROP CONSTRAINT IF EXISTS alarm_single_origin;`,
 				},
 			},
+			{
+				Id: "alarms_3",
+				Up: []string{
+					`ALTER TABLE alarms ADD COLUMN level  SMALLINT NOT NULL DEFAULT 1;`,
+					`ALTER TABLE alarms ADD COLUMN status VARCHAR(10) NOT NULL DEFAULT 'active';`,
```
manuio: status VARCHAR(10) NOT NULL DEFAULT 'active' CHECK (status IN ('active', 'noted', 'cleared'))

manuio (changes requested):

anna02272 (approved):

duolok (approved):
LGTM

manuio (approved):
LGTM

PR #1199, opened 2026-04-16
MF-1198 - Add alarm publish/subscribe and refactor consumer interfaces 

### LINKED ISSUE #1198: Add typed alarm publish/subscribe and refactor consumer interfaces
Several related problems existed across the messaging layer:

1. **No typed alarm support in the proto and NATS layer**
`mfx.proto` had no `Alarm` message type. The rules service had no way to publish alarms to NATS in a structured, typed way, and the alarms service had no way to receive them as typed events from the broker.

2. **Generic, type-unsafe consumer interface `consumers.Consumer ` used `any` as the message parameter**
Every service had to perform runtime type assertions inside `Consume`, which could fail at runtime for mismatches that should be caught at compile time. Services that only consume messages still need to satisfy the same interface as services that might consume alarms — violating interface segregation.
```
type Consumer interface {
    Consume(subject string, messages any) error
}
```
3. **`consumers.Start` was not extensible**
The signature `Start(id, sub, consumer, subjects...) ` assumed a single consumer type and a flat list of subjects. Supporting multiple message types (messages, alarms, future commands/notifications) would require either a bloated interface or multiple `Start` calls with no unified wiring point.

4. **Race condition in `pubsub.Subscribe`**
The re-subscribe logic in `Subscribe` called the public `Unsubscribe` method while holding the mutex, requiring an explicit unlock/re-lock around it. This left a window between releasing and re-acquiring the lock where another goroutine could modify the subscriptions map.

majabirmancevic:
Resolves #1198 :

`pkg/proto/mfx.proto + mfx.pb.go`
Added `Alarm` message with fields `thing_id`, `subtopic`, `protocol`, `created`, `level`, `rule_id`, `rule_info`(some of these fields will be used in this PR https://github.com/MainfluxLabs/mainflux/pull/1195). 
Rules service populates and publishes this; alarms service receives and persists it.

> Note: The `Payload` field is intentionally left out because it is planned to be discontinued due to unreadability. Instead, the `rule_info` field will contain all the necessary information.

`pkg/messaging/pubsub.go`
Added `AlarmPublisher`, `AlarmHandler`, `AlarmSubscriber` - broker-agnostic interfaces that mirror the existing `Publisher`/`MessageHandler`/`Subscriber`.

`pkg/messaging/nats/publisher.go`
Added `nats.Publisher` composite interface (`messaging.Publisher` + `messaging.AlarmPublisher`) and `PublishAlarm` implementation. The rules service accepts `nats.Publisher` since it needs both.

`pkg/messaging/nats/pubsub.go`
Added `nats.PubSub` composite interface (`messaging.PubSub` + `messaging.AlarmPublisher` + `messaging.AlarmSubscriber`)

- Added `SubscribeAlarms`/`UnsubscribeAlarms` with NATS wire format (protobuf unmarshal)

- Extracted internal subscribe/unsubscribe helpers (lock assumed held) — all four public methods now follow: validate --> lock --> defer unlock --> delegate to helper. Eliminates the previous race window where `Subscribe` released the lock to call `Unsubscribe` then re-acquired it.

`consumers/consumer.go`
- `Consumer.Consume(subject, any)` --> `MessageConsumer.ConsumeMessage(subject, protomfx.Message)` and `AlarmConsumer.ConsumeAlarm(subject, protomfx.Alarm)`

- `Start(id, sub, consumer, subjects...)` --> `Start(id, subs ...Subscription)` with `Messages(sub, consumer, subjects...)` and `Alarms(alarmSub, consumer)` constructors

- Each `Subscription` is a self-contained closure capturing its own subscriber 

All consumer services (`postgres-writer`, `timescale-writer`, `webhooks`, `smtp-notifier`, `smpp-notifier`, `alarms`, `rules` + their API middleware layers) updated to implement only the interface they need, with typed method signatures.


manuio (dismissed):
LGTM

pkg/messaging/nats/pubsub.go
```diff
@@ -158,14 +136,74 @@ func (ps *pubsub) Unsubscribe(id, topic string) error {
...
+		err error
+	)
+	if ps.queue != "" {
+		sub, err = ps.conn.QueueSubscribe(topic, ps.queue, nh)
+	} else {
+		sub, err = ps.conn.Subscribe(topic, nh)
+	}
+	if err != nil {
+		return err
+	}
```
duolok: Can we use a switch here?

rules/rules.go
```diff
@@ -60,24 +60,30 @@ func (rs *rulesService) processRule(msg *protomfx.Message, parsedPayload any, ru
...
-				subject = fmt.Sprintf("%s.%s.%s", subjectAlarms, alarms.AlarmOriginRule, rule.ID)
-			default:
-				continue
+		switch action.Type {
+		case ActionTypeSMTP, ActionTypeSMPP:
+			var subjectPrefix string
+			if action.Type == ActionTypeSMTP {
+				subjectPrefix = subjectSMTP
+			} else {
+				subjectPrefix = subjectSMPP
```
duolok: and here

manuio (approved):
LGTM

duolok (approved):
LGTM

pkg/messaging/pubsub.go
bool3max, line 84: Now that there's an `AlarmSubscriber` type in the same package, could this maybe be `MessageSubscriber` instead?

majabirmancevic: this is the base `Subscriber`, which is broker-agnostic and is also used in mqtt and rabbitmq so changing the name would touch many files

majabirmancevic: in order to maintain that consistency, we should also rename the publisher to MessagePublisher then, but as I said, it is also used in other brokers

bool3max (approved):
LGTM

PR #1203, opened 2026-04-21
MF-1202 - Add first and last aggregation types

### LINKED ISSUE #1202: Add first and last aggregation types
Add first and last aggregation types that return the earliest and latest complete row per time bucket respectively.
Useful for non-numeric payloads such as GPS coordinates or string fields where min/max/avg are meaningless.

manuio:
Resolves #1202

### DIFF
--- a/api/openapi/readers.yml
+++ b/api/openapi/readers.yml
@@ -613,7 +613,7 @@ components:
           default: 1
         agg_type:
           type: string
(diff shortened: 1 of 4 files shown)


duolok (approved):
LGTM

anna02272 (approved):

PR #1210, opened 2026-04-24
MF-1160 - Enhance MF Filestore, use SeaweedFS for storage

### LINKED ISSUE #1160: Rules: Dynamic subject subscription Enhance filestore: object storage + streaming for group files
# Context

The group files flow (POST /groups/:id/files, GET /groups/:id/files/:name, etc.) has the same storage and memory problems as thing files, plus sequential-only bulk deletion.

# Problems

Same local filesystem dependency and no storage abstraction
ViewGroupFile / ViewGroupFileByKey return []byte — OOM on large files
Silent DB error bug in postgres/groups.go:65 — checks err instead of errdb
RemoveAllFilesByGroup deletes files one-by-one sequentially (noted in code as needing parallelism)
No integrity verification on group-owned files

# Proposed Changes

Reuse the FileStore interface introduced in the thing files issue
Plug SeaweedFS backend into the group files paths
Stream file I/O for ViewGroupFile and ViewGroupFileByKey
Add SHA256 checksum on upload, verify on download
Fix the errdb bug in postgres/groups.go
Parallelize RemoveAllFilesByGroup using a worker pool

Depends on issue:

duolok:
Resolves #1160

This PR puts a pluggable object store
behind the service and adds SeaweedFS as a backend, so file contents live in
durable, shared storage while metadata stays in Postgres/Timescale.

**Storage abstraction**
- New `filestore/store` package with a small `FileStore` interface
  (`Put` / `Get` / `Delete` / `DeletePrefix`) and two backends:
  - `local` — filesystem (default, unchanged behaviour for dev)
  - `seaweedfs` — talks to a SeaweedFS filer over HTTP
- Backend is selected with `MF_FILESTORE_BACKEND`.

- Uploads compute a SHA256 checksum; downloads stream from the backend and
  verify it at end-of-stream. On a mismatch the transfer is aborted
  mid-stream (status/headers are already sent), so clients must treat a
  truncated download as a failure. A missing object for an existing row
  returns `404` instead of `500`.

**Downloads now stream** instead of buffering the whole file in memory, and
set an RFC 6266 `Content-Disposition` (with `filename*` for non-ASCII names).

**Upload hardening**
- Max upload size via `MF_FILESTORE_MAX_UPLOAD_MB` (also passed to the filer).
- Magic-byte sniffing: the file's real content type must match its extension
  (binaries/pointclouds/BIM are exempt).
- Filename validation: rejects path separators, `..`, control chars, invalid
  UTF-8, and over-long names.

### Config

| Env var | Default | Purpose |
|---|---|---|
| `MF_FILESTORE_BACKEND` | `local` | `local` or `seaweedfs` |
| `MF_FILESTORE_SEAWEED_URL` | `http://localhost:8888` | filer base URL |
| `MF_FILESTORE_SEAWEED_PREFIX` | `filestore` | key prefix on the filer |
| `MF_FILESTORE_SEAWEED_TIMEOUT` | `30s` | dial/TLS/response-header timeout |
| `MF_FILESTORE_MAX_UPLOAD_MB` | `1024` | max upload size |


filestore/postgres/init.go
```diff
@@ -78,6 +78,33 @@ func migrateDB(db *sqlx.DB) error {
 					`ALTER TABLE things_files DROP COLUMN group_id`,
 				},
 			},
+			{
+				Id: "filestore_3",
```
anna02272, line 82: Could these be combined into a single migration instead of three?

filestore/api/http/responses.go
```diff
@@ -81,6 +85,52 @@ func (res viewFileRes) Empty() bool {
...
+
+// contentDisposition builds an RFC 6266 attachment header. ASCII names are
+// emitted with a quoted filename; non-ASCII names additionally include a
+// filename* parameter per RFC 5987 so UTF-8 is conveyed faithfully.
+func contentDisposition(name string) string {
+	if isASCII(name) {
+		return fmt.Sprintf(`attachment; filename=%q`, name)
+	}
+	ascii := strings.Map(func(r rune) rune {
+		if r > 0x7f {
```
manuio: create a const for this magic number

filestore/api/http/transport.go
```diff
@@ -521,15 +574,74 @@ func getListFilesParams(r *http.Request) (listFilesParams, error) {
...
 // ParseFileName returns file class and format based on file name.
+// Rejects names with path separators, traversal sequences, control
+// characters, invalid UTF-8, or lengths beyond maxFileNameLen so that
+// downstream URL building, header emission, and DB writes are safe.
 func parseFileName(name string) (string, string, error) {
-	// Find the last dot in the filename
+	if name == "" || len(name) > maxFileNameLen {
+		return "", "", apiutil.ErrInvalidQueryParams
+	}
+	if strings.ContainsAny(name, `/\`) || strings.Contains(name, "..") || strings.HasPrefix(name, ".") {
```
manuio, line 638: Not sure here. Should we create consts?

filestore/api/http/transport.go
```diff
@@ -521,15 +574,74 @@ func getListFilesParams(r *http.Request) (listFilesParams, error) {
...
+		return "", "", apiutil.ErrInvalidQueryParams
+	}
+	if !utf8.ValidString(name) {
+		return "", "", apiutil.ErrInvalidQueryParams
+	}
+	for _, r := range name {
+		if r < 0x20 || r == 0x7f {
```
manuio: Magic numbers

filestore/store/local.go
```diff
@@ -0,0 +1,138 @@
...
+		}(k)
+	}
+	wg.Wait()
+	close(errCh)
+	var errs []error
+	for e := range errCh {
+		errs = append(errs, e)
+	}
+	return errors.Join(errs...)
+}
```
manuio: Add some blank lines to make it more readable

filestore/store/local.go
```diff
@@ -0,0 +1,138 @@
...
+
+func (l *local) DeletePrefix(_ context.Context, prefix string) error {
```
manuio, line 70: same here

manuio (changes requested):

filestore/api/http/transport.go
duolok: I wouldn't create consts for this. I think it's readable as is.

cmd/filestore/main.go
```diff
@@ -65,6 +66,19 @@ const (
 	defDBSSLKey          = ""
 	defDBSSLRootCert     = ""
 	defESURL             = "redis://localhost:6379/0"
+	defFilesPath         = "files"
+	envFilesPath         = "MF_FILESTORE_FILES_PATH"
```
majabirmancevic, line 70: move all `env` consts down

docker/docker-compose.yml
```diff
@@ -19,6 +19,9 @@ volumes:
   mainfluxlabs-timescale-volume:
   mainfluxlabs-filestore-db-volume:
   mainfluxlabs-filestore-volume:
+  mainfluxlabs-seaweedfs-master-volume:
+  mainfluxlabs-seaweedfs-volume-volume:
```
majabirmancevic, line 23: should we keep it like this `volume-volume`?

cmd/filestore/main.go
```diff
@@ -225,6 +245,26 @@ func loadConfig() config {
 		thingsConfig:      thingsConfig,
 		authConfig:        authConfig,
 		esURL:             mainflux.Env(envESURL, defESURL),
+		maxUploadBytes:    maxUploadMB * 1024 * 1024,
+	}
+}
+
+func buildStore(logger logger.Logger) store.FileStore {
```
anna02272: The ```logger logger.Logger```  is not used, both failure paths call ```log.Fatalf``` instead

filestore/README.md
```diff
@@ -1,6 +1,8 @@
...
+The Filestore service provides file storage for things and groups. File contents are written to a pluggable object store (local filesystem or SeaweedFS filer); file metadata is persisted to a database. Files can be scoped to an individual thing (authenticated with a thing key) or to a group (authenticated with a user bearer token at editor level or above).
+
+Group-file uploads compute a SHA256 checksum on ingest and store it in the database. Group-file downloads stream from the backend and verify the checksum at end-of-stream. Because the body is streamed, the `200 OK` status and headers are sent before verification completes: a mismatch (or a backend read failure) cannot change the status code, so the server logs the error and aborts the connection mid-transfer. Clients must therefore treat a truncated or aborted download as a failure rather than relying solely on the status code. A download whose object is missing entirely (index row present, object gone) returns `404`.
```
anna02272, line 5: Is this only for group files or thing files also? Should we say only ```File``` instead of ```Group-file```?

duolok: This is for both group and thing files.

filestore/store/local_test.go
```diff
@@ -0,0 +1,270 @@
...
+
+	"github.com/MainfluxLabs/mainflux/filestore/store"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	testKey     = "groups/g1/file.txt"
+	testContent = "hello mainflux"
+	wrongSum    = "deadbeef"
```
majabirmancevic, line 27: let's use a better example

duolok: What's wrong here?

majabirmancevic: since real checksums here are 64-character sha256 hex strings, then a 64-character hex string could be a better fit

duolok: I don't think that matters, the point it so be wrong

docker/docker-compose.yml
majabirmancevic: ^ @duolok?

filestore/api/http/transport.go
```diff
@@ -521,15 +578,79 @@ func getListFilesParams(r *http.Request) (listFilesParams, error) {
...
+func mimeMatches(class, format, mime string) bool {
+	switch class {
+	case binariesClass, pointcloudsClass, bimClass:
+		return true
+	case imagesClass:
+		// SVG is an XML document, not a binary image format, so
+		// http.DetectContentType never reports image/* for it: an XML prolog
+		// sniffs as text/xml and a bare <svg> root as text/plain.
+		if format == svgFormat {
+			return strings.HasPrefix(mime, "text/") || strings.HasPrefix(mime, "image/")
```
majabirmancevic: we could use consts for these prefixes

filestore/mocks/things.go
```diff
@@ -0,0 +1,228 @@
+// Copyright (c) Mainflux
+// SPDX-License-Identifier: Apache-2.0
+
+package mocks
```
majabirmancevic, line 4: separate this `repos.go` into `things.go` and `groups.go`

filestore/store/seaweedfs.go
```diff
@@ -0,0 +1,179 @@
...
+// connect, TLS handshake, and response-header phases; the overall request
+// deadline is driven by the caller's context so body transfers of large
+// objects are not capped by a single whole-request timeout.
+func NewSeaweedFS(rawURL, prefix string, timeout time.Duration) (FileStore, error) {
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		return nil, fmt.Errorf("parse seaweedfs url: %w", err)
+	}
+
+	if u.Scheme != "http" && u.Scheme != "https" {
```
majabirmancevic: and probably here -> consts

filestore/mocks/things.go
duolok: done

filestore/mocks/things.go
```diff
@@ -0,0 +1,131 @@
...
+
+func NewThingsRepository() *ThingsRepository {
+	return &ThingsRepository{
+		byKey: map[string]filestore.FileInfo{},
+		group: map[string]string{},
+	}
+}
+
+func thingKey(thingID string, fi filestore.FileInfo) string {
+	return thingID + "|" + fi.Class + "|" + fi.Format + "|" + fi.Name
```
majabirmancevic: "|" or "/"?

filestore/store/seaweedfs.go
```diff
@@ -0,0 +1,184 @@
...
+// connect, TLS handshake, and response-header phases; the overall request
+// deadline is driven by the caller's context so body transfers of large
+// objects are not capped by a single whole-request timeout.
+func NewSeaweedFS(rawURL, prefix string, timeout time.Duration) (FileStore, error) {
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		return nil, fmt.Errorf("parse seaweedfs url: %w", err)
+	}
+
+	if u.Scheme != httpScheme && u.Scheme != httpsScheme {
```
majabirmancevic, line 48: should we take care of this?

duolok: ?

majabirmancevic: won't we get an error if it's an invalid value? why do we have to check?

duolok: I don't see why wouldn't we want this

majabirmancevic: no need for an additional check if it will be verified later, upon receipt of the request?

duolok: No, because this is an URL used for connecting to seaweed, not URL for hitting an endpoint

PR #1211, opened 2026-04-24
MF-1218 - refactor Redis event stream code

### LINKED ISSUE #1218: Event stream implementation issues
Briefly, the current implementation of Redis streams suffers from multiple issues:

* The code structure and types/interfaces for much of the event-related code doesn't make sense
* Event consumers have no way to safely recover from a crash/timeout and they lose all events emitted in the meantime
* The error handling of event consumers is completely broken: they consume events in batches, but discard the rest of the batch the moment handling one of the events in a batch fails
* **Event emitters completely lack any error handling at** all - they silently discard Redis `XADD` errors without even logging them. Moreover, there's no retry logic at all - the moment emitting a certain event fails for whatever reason (e.g. Redis is down, network issues, ...), it is lost forever

bool3max:
* Revamped the existing types and interfaces of event-related code which were quite nonsensical and confusing    
* `EventHandler` now receives concrete, structured event types. Those types, as well as the functions for decoding raw Redis messages into them are now defined only once in the `pkg/events` package, instead of being duplicated across every interested consumer service's local event handling code
* Migrated from Redis consumer groups to pure `XADD/XREAD` with a single consumer per service as we do not currently support any kind of horizontal scaling between services. This comes with a number of fixes to event handling:
    * Consumers now properly recover after crashes and no longer miss/skip events. The previous consumer group implementation didn't take advantage of the relevant tools provided to handle this (`XPENDING`,`XCLAIM`, the PEL, etc...). With this simpler single-consumer implementation, each consumer keeps track of the last-received message ID in the redis event store itself and can as such properly recover after any amount of downtime. 
    * Consumers now properly deal with errors occurring during `EventHandler` operation. Previously, a failure during the handling of a single event in a batch of Redis messages **skipped the rest of the batch entirely**. Now, failure is handled by retrying the offending event at least once + logging, and only then continuing with the rest of the batch.
    * Failures in the Redis `Subscriber` `XREAD` loop are now handled with an exponential backoff retry with a delay of up to 30sec.
* Introduced a new `pkg/events.Publisher` interface and implementation, which brings multiple improvements and fixes over the way events were previously published from middlewares:
    + Service-specific event store middlewares now wrap this event publisher. As such, they no longer have to care about Redis-specific logic or types (`XAdd` and its various options, error handling, etc...). To publish an event, they simply call `.Publish()`, passing an instance of a concrete typed event, which makes event middleware methods much cleaner.
    + Proper error handling. Previously, errors during an `XADD` of an event in a middleware method were always completely ignored, they weren't even being logged. This potentially led to awkward states where the destructive operation that resulted in the event was persisted, but, as the event was never retried/re-fired, interested consumers never received it, and so the state between services drifted.
        + The new publisher addresses this by instead **maintaining an in-memory buffer of queued events**, which is continuously drained by a background goroutine that `XADD`s them to Redis.
        + `Publish()` is asynchronous, as it simply enqueues the event onto the buffer channel. 
            + Note also that, in contrast, previously, each middleware method synchronously called `XADD` which blocked the incoming HTTP/gRPC request that triggered it, yet it swallowed the error
            + If the buffer is full, the oldest event is dropped to make space
        + The drainer goroutine receives events from the buffer channel and emits them to Redis. It handles errors with an exponential backoff sleep of up to 30 seconds + a log. Failed events remain in the queue and are continuously retried in order, instead of being dropped. In the meantime, new events can safely be enqueued by the publisher.
    + Note that, while substantially better than what we had before, even this implementation of the publisher isn't perfect. For example, a crash of the service process emitting the events can potentially lead to full loss of queued events, as they are stored in-memory. From what I've seen, the safest and most reliable way to protect against crashes and outages is to persist events to a database in the same DB transaction as the operation they're associated with (e.g. Thing removal), while having a background job continuously listening for, consuming, and emitting the events. 

### TODO:

* [x] Properly handle Redis `XADD` errors that occur in event producers. Currently, if an event producer carries out a destructive action (e.g. deletion of Things), but sending out the associated redis event _fails_, it isn't retried, and the destructive operation is not rolled back, so the state between services drifts.


docker/.env
```diff
@@ -61,6 +61,8 @@ MF_AUTH_DB=auth
 MF_AUTH_SECRET=secret
 MF_AUTH_LOGIN_TOKEN_DURATION=10h
 MF_AUTH_ES_URL=redis://localhost:6379/0
+MF_AUTH_ES_STREAM_MAXLEN=100000
```
majabirmancevic, line 64: maybe `MAX_LEN` to be consistent with other envs

bool3max: For `things` it's `MF_THINGS_ES_STREAM_MAXLEN` as well. I think `MAXLEN` is fine

pkg/events/publisher.go
```diff
@@ -0,0 +1,292 @@
...
+	// DefStreamMaxLen is the default approx number of messages in the Redis event stream.
+	DefStreamMaxLen = 100000
+
+	// DefBufferSize is the default capacity of the publish
+	// buffer channel.
+	DefBufferSize = 10000
+
+	// DefDrainIntervalInitial is the initial sleep duration between XADD retries when
+	// emitting an event fails.
+	DefDrainIntervalInitial = time.Second
```
majabirmancevic: an adjective should come before a noun, right? `DefDrainInitialInterval`

pkg/events/publisher.go
```diff
@@ -0,0 +1,292 @@
...
+
+	// DefDrainBackoffMax is the maximum backoff duration between XADD retries.
+	DefDrainBackoffMax = 30 * time.Second
```
majabirmancevic: DefDrainMaxBackoff

docker/.env
majabirmancevic: I thought that since it's considered two words, we could separate them

bool3max: I don't think coupling these is that uncommon. Besides the redis option that's set by these two is `MAXLEN` too.

cmd/auth/main.go
```diff
@@ -170,8 +170,8 @@ func main() {
 		Stream:               mfevents.AuthStream,
 		MaxLen:               cfg.esStreamMaxLen,
 		BufferSize:           cfg.esBufferSize,
-		DrainIntervalInitial: mfevents.DefDrainIntervalInitial,
-		DrainBackoffMax:      mfevents.DefDrainBackoffMax,
+		DrainIntervalInitial: mfevents.DefDrainInitialInterval,
+		DrainBackoffMax:      mfevents.DefDrainMaxBackoff,
```
majabirmancevic:
```
DrainIntervalInitial:
DrainBackoffMax:
```
should we fix it in the `PublisherConfig` as well?

cmd/auth/main.go
```diff
@@ -159,7 +165,24 @@ func main() {
...
 	defer dbCloser.Close()
 
-	esClient := connectToRedis(cfg.esURL, logger)
+	pub, err := mfevents.NewPublisher(mfevents.PublisherConfig{
+		URL:                  cfg.esURL,
+		Stream:               mfevents.AuthStream,
+		MaxLen:               cfg.esStreamMaxLen,
+		BufferSize:           cfg.esBufferSize,
+		DrainInitialInterval: mfevents.DefDrainInitialInterval,
+		DrainMaxBackoff:      mfevents.DefDrainMaxBackoff,
```
majabirmancevic, line 174: does it make sense to make this configurable with envs too?

bool3max: Good question but I don't these two need to be configurable at runtime.

pkg/events/redis.go
```diff
@@ -32,34 +33,31 @@ const (
 	AuthStream   = mainfluxPrefix + "auth"
 )
```
duolok, line 35: Could some of these consts be removed? They were only used in the `*/events.go` files, which will no longer exist.

duolok: Or are we just missing the corresponding `handler.go` implementations for those?

bool3max: Individual event handlers now receive concrete event types, they no longer have to check/match the operation strings i.e. the consts above. The consts are still however used by:

1. `decodeEvent()` in `pkg/events/schema.go` to decode a raw RedisEvent into its typed counterpart
2. The concrete typed event's `.Encode()` method which turns it into the generic RedisEvent which carries the operation string

duolok (dismissed):
LGTM

manuio (approved):
LGTM

majabirmancevic (approved):

duolok (approved):
LGTM

PR #1212, opened 2026-04-27
MF-1197 - Introduce Input field and refactor assign/unassign endpoints

### LINKED ISSUE #1197: Rules: introduce typed Input model and replace per-thing assignment
## Context

Rules are currently assigned to individual things via a separate `rules_things` join table. This means the same rule must be manually assigned to every device in a group — which doesn't scale and is inconsistent with how every major IoT platform handles rule scoping.

Beyond the assignment model, the current `Rule` struct has no concept of an input source. `Consume()` in `service.go` hardcodes a single entry point that only handles `protomfx.Message` events, making it impossible to trigger rules from other event sources such as alarms, commands, or schedules.

This issue introduces the `Input` model — a typed, explicit declaration of what event source a rule listens to and which things within the group it applies to. This is the foundational change that all subsequent input type issues depend on.

**Input types (defined as constants in rules/rules.go)**
- `message`  — triggered by sensor data published by a thing
- `alarm`    — triggered by an alarm at a given severity
- `command`  — triggered when a command is dispatched to a thing

## Changes

**rules/rules.go**
- Add `Input` struct with `Type string` and `ThingIDs []string`
- Add `Input` field to `Rule`

**rules/service.go**
- Remove `ServiceRules`: `AssignRules`, `UnassignRules`, `UnassignRulesByThing`
- Remove `RepositoryRules`: `Assign`, `Unassign`, `UnassignByThing`
- Refactor `ListRulesByThing` to filter `RetrieveByGroup` results by `Input.ThingIDs contains thingID`
- Refactor `ListThingIDsByRule` to derive from `Input.ThingIDs` instead of join table
- `Consume()`: replace `RetrieveByThing(msg.Publisher)` with `RetrieveByGroup(groupID)`, filter in-memory by rules whose `Input.ThingIDs` contains `msg.Publisher`

**rules/api/http/rules/transport.go**
- Remove `POST /things/:id/rules`
- Remove `PATCH /things/:id/rules`

**rules/api/http/rules/endpoint.go**
- Remove `assignRulesEndpoint`, `unassignRulesEndpoint`

**rules/api/http/rules/requests.go**
- Remove `thingRulesReq`

**rules/api/logging.go, rules/api/metrics.go, rules/tracing/rules.go, rules/mocks/rules.go**
- Remove all corresponding method implementations for removed methods only

**rules/postgres/rules.go**
- Remove `Assign`, `Unassign`, `UnassignByThing`
- Refactor `RetrieveByThing` and `RetrieveThingIDsByRule` to query against `input` JSONB column

**rules/postgres/init.go**
- New migration: drop `rules_things` table, add `input` JSONB column to `rules`

## Acceptance Criteria

- `thing_ids` is required on every input
- `Consume()` evaluates only rules whose `Input.ThingIDs` contains the publishing thing
- `GET /things/:id/rules` and `GET /rules/:id/things` continue to work via JSONB queries
- No per-thing assignment endpoints remain

majabirmancevic:
Closes #1197

This PR introduces the `Input` model for rules, which explicitly defines the event source type and the list of thing IDs a rule applies to. It also refactors the assign/unassign endpoints.

### What was implemented

- Added `input_type` column to the `rules` table with an index on `(group_id, input_type)` for efficient filtering
- `thing_ids` are required when **creating** a rule — a rule without any assigned things has no meaningful trigger
- Retained and refactored assign/unassign endpoints (`POST /rules/:id/things`, `PATCH /rules/:id/things`) for managing rule-thing relationships after creation; added an index on `rules_things(thing_id)` for efficient lookups
- Extracted shared validation helpers (`validateThingIDs`, `validateInputType`, `validateConditions`, `validateActions`) to eliminate duplication between create and update request validation
- Update request (`PUT /rules/:id`) uses a dedicated `updateRuleInput` struct that does not accept `thing_ids`, keeping rule property updates and thing assignment as separate operations

### DIFF
--- a/api/openapi/rules.yml
+++ b/api/openapi/rules.yml
@@ -16,7 +16,7 @@ paths:
       requestBody:
         $ref: "#/components/requestBodies/CreateRulesReq"
       responses:
-        '201':
+        '201': 
           $ref: "#/components/responses/CreateRulesRes"
         '400':
           description: Failed due to malformed JSON.
@@ -35,6 +35,12 @@ paths:
         - rules
       parameters:
         - $ref: "#/components/parameters/GroupId"
+        - $ref: "#/components/parameters/Limit"
+        - $ref: "#/components/parameters/Offset"
+        - $ref: "#/components/parameters/Order"
+        - $ref: "#/components/parameters/Dir"
+        - $ref: "#/components/parameters/Name"
+        - $ref: "#/components/parameters/InputType"
       responses:
         '200':
           $ref: "#/components/responses/ListRulesRes"
@@ -56,6 +62,12 @@ paths:
         - rules
       parameters:
         - $ref: "#/components/parameters/ThingId"
+        - $ref: "#/components/parameters/Limit"
+        - $ref: "#/components/parameters/Offset"
+        - $ref: "#/components/parameters/Order"
+        - $ref: "#/components/parameters/Dir"
+        - $ref: "#/components/parameters/Name"
+        - $ref: "#/components/parameters/InputType"
       responses:
         '200':
           $ref: "#/components/responses/ListRulesRes"
@@ -67,54 +79,6 @@ paths:
           description: Failed to perform authorization over the entity.
         '422':
           description: Database can't process request.
-        '500':
-          $ref: "#/components/responses/ServiceError"
-    post:
-      summary: Assigns rules to a thing
-      description: Assigns rules to a specific thing identified by the provided ID.
-      tags:
-        - rules
-      parameters:
-        - $ref: "#/components/parameters/ThingId"
-      requestBody:
-        $ref: "#/components/requestBodies/ThingRulesReq"
-      responses:
-        '200':
-          description: Rules assigned successfully.
-        '400':
-          description: Failed due to malformed JSON.
-        '401':
-          description: Missing or invalid access token provided.
-        '403':
-          description: Failed to perform authorization over the entity.
-        '404':
-          description: Thing or one of the rules does not exist.
-        '415':
-          description: Missing or invalid content type.
-        '500':
-          $ref: "#/components/responses/ServiceError"
-    patch:
-      summary: Unassigns rules from a thing
-      description: Unassigns rules from a specific thing identified by the provided ID.
-      tags:
-        - rules
-      parameters:
-        - $ref: "#/components/parameters/ThingId"
-      requestBody:
-        $ref: "#/components/requestBodies/ThingRulesReq"
-      responses:
-        '200':
-          description: Rules unassigned successfully.
-        '400':
-          description: Failed due to malformed JSON.
-        '401':
-          description: Missing or invalid access token provided.
-        '403':
-          description: Failed to perform authorization over the entity.
-        '404':
-          description: Thing or one of the rules does not exist.
-        '415':
-          description: Missing or invalid content type.
         '500':
           $ref: "#/components/responses/ServiceError"
   /rules/{ruleId}/things:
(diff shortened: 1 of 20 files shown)


majabirmancevic:
### Deviations from the original issue

#### 1. `input` stored as separate columns instead of JSONB

The issue proposes storing the entire `input` object (type + thing_ids) as a single JSONB column and querying it with JSONB operators.

We instead store `input_type` as a `TEXT` column and keep `thing_ids` in the existing `rules_things` join table. Reasons:

- **Query performance** — filtering by `thing_id` inside a JSONB array requires `@>` or `json_array_elements` operators, which are significantly slower than an indexed foreign key join, especially as the number of rules and things grows
- **Data integrity** — the join table enforces UUID type correctness and supports `ON DELETE CASCADE`, which JSONB cannot provide natively
- **Atomic updates** — adding or removing a single thing from a JSONB array is not atomic without explicit locking (read-modify-write), whereas the join table handles individual inserts and deletes atomically
- **Indexing** — `idx_rules_things_thing_id` is a lightweight B-tree index; an equivalent GIN index on a JSONB array is heavier and less precise for this access pattern

#### 2. Assign/unassign endpoints retained instead of removed

The issue proposes eliminating the assign/unassign endpoints in favor of always passing `thing_ids` through the rule payload.

We kept the endpoints because requiring `thing_ids` on the update endpoint introduces a practical usability problem: to add a single thing to a rule that already has ten assigned things, a client would need to know and re-send all ten existing IDs plus the new one. This is error-prone and makes the update non-idempotent in practice. The dedicated assign/unassign endpoints allow targeted, atomic changes to the rule-thing relationship without affecting rule properties.

`thing_ids` remain required on create as the initial assignment, which is a natural part of rule setup.

rules/api/http/rules/responses.go
```diff
@@ -88,6 +88,21 @@ func (res RulesPageRes) Empty() bool {
 	return false
 }
 
+// TODO: Consider introducing apiutil.EmptyRes with a configurable status code.
```
majabirmancevic: ^ _question:_ check it to open an issue 

(and remove all TODO comments)

rules/postgres/rules.go
```diff
@@ -307,19 +295,47 @@ func (rr ruleRepository) retrieveRules(ctx context.Context, query, cquery string
...
+	}
+	return ids, nil
+}
+
+func insertRulesThings(ctx context.Context, tx *sqlx.Tx, ruleID string, thingIDs []string) error {
+	if len(thingIDs) == 0 {
+		return nil
+	}
+
+	// TODO: replace with a single bulk INSERT for better performance.
```
majabirmancevic: ^ for this too

rules/service.go
```diff
@@ -252,83 +261,64 @@ func (rs *rulesService) UpdateRule(ctx context.Context, token string, rule Rule)
...
-	grID, err := rs.things.GetGroupIDByThing(ctx, thingID)
-	if err != nil {
+	if err := rs.things.CanUserAccessGroup(ctx, domain.UserAccessReq{Token: token, ID: rule.GroupID, Action: domain.GroupEditor}); err != nil {
 		return err
 	}
-
-	for _, id := range ruleIDs {
-		rule, err := rs.rules.RetrieveByID(ctx, id)
+	// TODO: replace with a single bulk gRPC call, e.g. CanUserAccessThings(ctx, groupID, thingIDs...)
+	// to avoid N round-trips to the things service.
```
majabirmancevic: ^ and this

rules/api/http/rules/responses.go
manuio: Create an issue for this one

majabirmancevic: here it is: https://github.com/MainfluxLabs/mainflux/issues/1221

rules/service.go
manuio: Right, let's create an issue.

majabirmancevic: related issue: https://github.com/MainfluxLabs/mainflux/issues/1222

rules/postgres/rules.go
majabirmancevic: related issue: https://github.com/MainfluxLabs/mainflux/issues/1224

api/openapi/rules.yml
```diff
@@ -16,7 +16,7 @@ paths:
       requestBody:
         $ref: "#/components/requestBodies/CreateRulesReq"
       responses:
-        '201':
+        '201':
```
manuio: Why this white space?

rules/postgres/rules.go
```diff
@@ -341,12 +356,22 @@ func toDBRule(r rules.Rule) (dbRule, error) {
...
 
+func toRuleThings(dbr dbRule, thingIDs []string) (rules.Rule, error) {
+	r, err := toRule(dbr)
+	if err != nil {
+		return rules.Rule{}, err
+	}
+	r.Input.ThingIDs = thingIDs
+	return r, nil
+}
+
```
bool3max: Is this necessary? Could the existing `toRule()` helper perhaps just accept `thingIDs` as a separate argument?

majabirmancevic: I separated them because "list" responses don't retrieve thing_ids, but yes, we could pass "nil" as an argument in `retrieveRules`, it's not that dramatic

anna02272 (approved):

duolok (approved):
LGTM

manuio (approved):
LGTM

PR #1214, opened 2026-04-27
MF-1213 - Add SUM aggregation support for time-series queries

### LINKED ISSUE #1213: Add SUM aggregation support for time-series queries
Add sum aggregation type to calculate the total sum of numeric field values within each time bucket, similar to how count works but summing values instead of counting messages. This should be supported for both SenML (value field) and JSON (numeric fields via agg_fields) message formats.

manuio:
Resolves #1213


majabirmancevic (approved):

bool3max (approved):
LGTM

PR #1216, opened 2026-04-29
MF-1215 - Return null for missing aggregation fields

### LINKED ISSUE #1215: Return null for missing aggregation fields in messages buckets
When aggregating JSON messages over time intervals, buckets containing records that lack the requested aggregation field are silently dropped from the response. Missing buckets should appear with a null value instead.

manuio:
Resolves #1215


duolok (approved):
LGTM

majabirmancevic (approved):

PR #1229, opened 2026-05-25
MF-1232 - Attach actor identity information to published events

### LINKED ISSUE #1232: Accompany platform events with actor identity information
As a prerequisite to introducing the audit log service, it's necessary that platform events be accompanied by an identity of their actor, i.e. the user who initiated the event. For example, if a "thing removed" event is emitted by the things service, we want to know who removed the thing.

bool3max:
### Background

As a prerequisite to introducing the audit log service, it's necessary that platform events be accompanied by an identity of their _actor_, i.e. the user who initiated the event. For example, if a "thing removed" event is emitted by the `things` service, we want to know _who_ removed the thing.

### JWT token identification

To attach actor identity (represented by the `domain.Identity{Email string, ID string}` type), it's necessary to decode the authentication JWT. This is usually done by the `auth` service `Identify` method, which takes a JWT and returns back a `domain.Identity` representing the exact user identified by the token.

When thinking about _where_ to perform JWT identification in the event publishing pipeline, I initially gravitated towards the event publisher `.Publish()` method : make it accept a token and call out to `auth.Identify`, then attach the identity to the event before emitting it.

The key issue with that approach is that _most_ service methods that accept an auth token as an argument __already decode it (i.e. pass it to `auth.Identify`) at some point during their execution__, most commonly to perform authorization/permission checks. As such, performing identification at the event publisher level would introduce an additional, potentially unneeded cross-service gRPC call.

I thus needed to find a way to decode the auth token into a `domain.Identity` only _once per endpoint_ and make it subsequently available to the service method itself as well as the event publishing middleware. 

The natural solution for this, partially implemented by this PR, is a __`go-kit` middleware__ - see [`pkg/authn.IdentityMiddleware`](https://github.com/bool3max/mainflux/blob/a52788eba0cc3e357e6948e1f1501fd8fef7bd65/pkg/authn/transport.go#L26): it extracts the auth token from the context, passes it to `auth.Identify()`, and then attaches the returned `domain.Identity` back to the context - making it accessible to all downstream code, including the event publisher, which can make use of it without needing to re-identify the token.

Note however, that in its current state, this PR does not solve the duplicate token identification calls. It merely lays the groundwork for a proper refactor at a later time, which would entail rewriting all service method definitions to access the token from the context as opposed to an argument, and rewriting all authorization/permission checks to accept/work with a `domain.Identity` as opposed to a raw token.

Currently, __only the event publisher makes use of the `domain.Identity` value from the context__. The rest of the code still follows the existing pattern where the token is extracted by a `go-kit` `DecodeRequestFunc` and then forwarded to the service methods (as an argument), after which they eventually perform `auth.Identify` calls themselves. As such, the `IdentityMiddleware` does not abort the request on error (e.g. missing auth token or a failure during `auth.Identify`), it simply logs the error. This is something that would be changed in the future.

---

> After inspecting the codebase for calls to `auth.Identify` I discovered some _existing_ code-paths that make duplicate calls to `auth.Identify`. For example, the `canUserAccessGroup` helper in `things`:
> https://github.com/MainfluxLabs/mainflux/blob/6cddb5d118f6a6b859f1508e8a2907c08cb89e5d/things/groups.go#L317-L336
> ... it first makes a call to `auth.Identify` to obtain the User ID, but then later potentially a call to `isAdmin` - which calls `auth.Authorize`, which internally calls `Identify()` again, because the `AuthzReq` includes a `token` opposed to a `domain.Identity`.
> So perhaps this is something that could be improved as well. My initial hunch is that all authz-related code should accept an _identity_ as opposed to a token.

### Other changes
- Renamed `pkg/auth` to `pkg/authn`
- Un-exported `pkg/events.RedisEvent` and its methods
- Made slight improvements to the structure of event-related types
    - `EventAction` types now define the operation string in a separate `Operation()` method, meaning that they no longer have to encode in manually in `Encode()`. Instead, the base `Event` type now features an `.Encode()` method which takes care of encoding both the operation and actor identity parameters into the final `redisEvent`


auth/api/http/keys/transport.go
```diff
@@ -12,44 +12,53 @@ import (
...
 	opts := []kithttp.ServerOption{
 		kithttp.ServerErrorEncoder(apiutil.LoggingErrorEncoder(logger, encodeError)),
+		kithttp.ServerBefore(authn.HTTPTokenToContext),
 	}
-	mux.Post("/keys", kithttp.NewServer(
-		kitot.TraceServer(tracer, "issue")(issueEndpoint(svc)),
+
+	withIdentity := authn.IdentityMiddleware(ac, logger)
+
+	newServer := func(name string, e endpoint.Endpoint, decodeFunc kithttp.DecodeRequestFunc) *kithttp.Server {
```
majabirmancevic: does it make sense to separate this into a helper since it is used in every service?

duolok: I'd also probably consider this instead of the approach in PR.

bool3max: Right - though another thing just occurred to me, and it's that I'm attaching the `IdentityMiddleware` to virtually every single endpoint, even though not all of them accept an auth token. Let me try and come up with a better pattern for this

majabirmancevic: that's what I wanted to ask you too, does it affect, since ThingKey is used somewhere for auth

bool3max: It doesn't affect it (the middleware would just see that there's no token and continue), but it's useless work being done for no reason, and it would become an issue once we fully make the switch to using the identity from the context properly

bool3max: I've removed the `newServer` factory in favor of just manually constructing the server for every endpoint, just as before. Endpoints that require authentication attach the `IdentityMiddleware` inside a call to `endpoint.Chain()`.

cmd/auth/main.go
```diff
@@ -200,10 +204,18 @@ func main() {
 
 	tc := thingsapi.NewClient(thConn, thingsTracer, cfg.timeout)
 
+	authConn := clientsgrpc.Connect(cfg.authConfig, logger)
+	defer authConn.Close()
+
+	authClientTracer, authClientCloser := jaeger.Init("auth_client", cfg.jaegerURL, logger)
```
majabirmancevic: `jaeger.Init("auth_http",...`

majabirmancevic: or auth_grpc?

cmd/converters/main.go
```diff
@@ -93,6 +100,15 @@ func main() {
 	defer pub.Close()
 
 	tc := thingsapi.NewClient(conn, thingsTracer, cfg.thingsGRPCTimeout)
+
+	authTracer, authCloser := jaeger.Init(svcName+"_auth", cfg.jaegerURL, logger)
```
majabirmancevic: let's be consistent with other services on this `svcName+"_auth"`

pkg/events/redis.go
```diff
@@ -29,28 +32,57 @@ const (
 	OrgCreate = orgPrefix + "create"
 	OrgRemove = orgPrefix + "remove"
 
+	actorIdentityUserID = "actor_identity_user_id"
```
majabirmancevic: I'd rather just use `user_id`/`user_email`, that's already clear?

and `actor` -> `user` naming

cmd/auth/main.go
bool3max: `auth` actually shouldn't need a gRPC client to itself - I've fixed it in the latest commit by making the `IdentityMiddleware` accept an interface with just the `Identify()` method instead of the entire `domain.AuthClient`. Now `auth` HTTP handlers pass the service instance when constructing the middleware, and other services pass the auth gRPC client, which also satisfies the interface

pkg/events/redis.go
bool3max: Perhaps `actor_identity_...` is too much but just `user_id` is likely to clash with the fields of some real events. I think `_actor_user_id` and `_actor_user_email` would be a nice middle ground

manuio: I'm not sure to udnerstand here. Can you give us an example of clash?

bool3max: Suppose there's a specific event that encodes one of its fields as `"user_id"` - that'd be a conflict. For example imagine that `users` emitted an event whenever a new user is registered, and that it encoded the ID of the user under the `"user_id"` field.

majabirmancevic: in that case, it should be just ID ("id") field, like we have everywhere
```
type ProfileUpdated struct {
	ID       string
	Name     string
	..
```

bool3max: We can't know what kind of fields future events may have. What if there's an event that needs both an ID and a "user ID"? With `_actor_user_id` (and `_actor_user_email`), we're pretty much totally sure that no fields of any future events will cause a conflict

majabirmancevic: so just because of one service, we have to use a more specific field everywhere? IMO, it makes more sense to use a specific field for `users` service when needed

bool3max: The field is shared by every event across all services, I think it makes sense for it to be as specific as possible. 

Besides, you'll hardly ever be looking at this field / its name directly, unless you for example attach to the redis instance and look at the raw events and their fields, and even then, I think `_actor_user_id` communicates the field's meaning a lot better than `user_id` for example. The `_actor` prefix at first glance tells me that this is something more deliberate and "private" to the platform.

duolok (approved):
LGTM

pkg/events/redis.go
majabirmancevic: "actor" naming sounds pretty inconsistent, is there a "triggered_by" option or something?

Issue #1230, opened 2026-05-27
Add profile dispatcher config flags for subject-based message routing


majabirmancevic:
Currently, all messages published by things are broadcast to `things.*.messages`, which is consumed by all services (rules, webhooks, writers) regardless of whether a given thing is configured to use them. This causes unnecessary processing — for example, the `webhooks` service performs a DB lookup for every message even if the publishing thing has no webhooks configured.

This issue introduces per-profile dispatcher flags that control which NATS subjects a message is published to at publish time, enabling selective routing and eliminating unnecessary service invocations.

Add 3 flags to ProfileConfig:
`webhook_enabled` — publish to webhooks subject (unconditional webhook forwarding / integrations)
`rule_enabled` — publish to a dedicated rules subject for rule evaluation
`write_enabled` — publish to `things.*.messages` for writers (storage)

PR #1231, opened 2026-05-27
MF-1165 - Add webhook as rule action

### LINKED ISSUE #1165: Support Webhook as Rule Action
Webhooks currently subscribe to `things.*.messages` and forward unconditionally per thing. In addition, it is necessary to enable the Webhook to be defined as a Rule Action. In that case, when a specific condition is met, the message is forwarded according to the defined Webhook configuration (URL, etc.).
Use the appropriate NATS subjects to distinguish them.

majabirmancevic:
Resolves #1165

### DIFF
--- a/cmd/webhooks/main.go
+++ b/cmd/webhooks/main.go
@@ -142,7 +142,7 @@ func main() {
 		return subscribeToThingsES(ctx, svc, cfg, logger)
 	})
 
-	if err = consumers.Start(svcName, consumers.Messages(pubSub, svc, nats.SubjectMessages)); err != nil {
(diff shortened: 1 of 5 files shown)


majabirmancevic:
> _Note:_ In this PR, the webhook service only subscribes to `nats.SubjectWebhooks` (due to rule-based actions), removing the ability to consume messages for integrations (current webhooks). This is a temporary state and will be resolved in the next PR. (related to #1230)

duolok (approved):
LGTM

manuio (approved):
LGTM

PR #1233, opened 2026-05-28
MF-1230 - Add flags in Profile Config

### LINKED ISSUE #1230: Add profile dispatcher config flags for subject-based message routing

majabirmancevic:
Resolves #1230


coap/adapter.go
```diff
@@ -65,8 +65,12 @@ func (svc *adapterService) Publish(ctx context.Context, key domain.ThingKey, msg
 		return err
 	}
 
-	if err := svc.pubsub.Publish(nats.GetMessagesSubject(msg.Publisher, msg.Subtopic), msg); err != nil {
-		return err
+	if pc.ProfileConfig != nil {
```
manuio: Should we move this check?

mqtt/handler.go
```diff
@@ -234,9 +234,18 @@ func (h *handler) Publish(c *session.Client, topic *string, payload *[]byte) {
 		return
 	}
 
-	if err := h.publisher.Publish(subject, msg); err != nil {
-		h.logger.Error(errors.Wrap(messaging.ErrPublishMessage, err).Error())
-		return
+	if isCmd {
```
manuio: switch case

manuio (changes requested):

mqtt/handler.go
majabirmancevic: replaced by an early return

manuio (approved):
LGTM

bool3max (approved):

PR #1235, opened 2026-06-03
NOISSUE - rename pkg/auth to pkg/authn

bool3max:
`pkg/auth` deals only with authe**n**tication-related code, as such I've renamed it to `pkg/authn`, as when other packages import it as `auth` it could be confused with the top-level `auth` service package


duolok (approved):
LGTM

majabirmancevic (approved):

manuio (approved):
LGTM

PR #1236, opened 2026-06-03
MF-1232 - attach user identity information to contexts originating from HTTP requests

bool3max:
### Background

As a prerequisite to introducing the audit log service, it's necessary that platform events be accompanied by an identity of their _actor_, i.e. the user who initiated the event. For example, if a "thing removed" event is emitted by the `things` service, we want to know _who_ removed the thing.

### JWT token identification

To attach actor identity (represented by the `domain.Identity{Email string, ID string}` type) to an event before it is dispatched, it's necessary to decode the authentication JWT associated with the HTTP request that led to the creation of the event. This is usually done by the `auth` service `Identify` method, which takes a JWT and returns back the said `domain.Identity` representing the exact user identified by the token.

When thinking about _where_ to perform JWT identification in the event publishing pipeline, I initially gravitated towards the event publisher `.Publish()` method : make it accept a token and call out to `auth.Identify`, then attach the identity to the event before emitting it.

The key issue with that approach is that _most_ service methods that accept an auth token as an argument __already decode it (i.e. pass it to `auth.Identify`) at some point during their execution__, most commonly to perform authorization/permission checks. As such, performing identification at the event publisher level would introduce an additional, potentially unneeded cross-service gRPC call.

I thus needed to find a way to decode the auth token into a `domain.Identity` only _once per endpoint_ and make it subsequently available to the service method itself as well as the event publishing middleware. 

The natural solution for this, partially implemented by this PR, is a __`go-kit` middleware__ - see [`pkg/authn.IdentityMiddleware`](https://github.com/bool3max/mainflux/blob/a52788eba0cc3e357e6948e1f1501fd8fef7bd65/pkg/authn/transport.go#L26): it extracts the auth token from the context, passes it to `auth.Identify()`, and then attaches the returned `domain.Identity` back to the context - making it accessible to all downstream code, including the event publisher, which can make use of it without needing to re-identify the token.

Note however, that in its current state, this PR does not solve the duplicate token identification calls. It merely lays the groundwork for a proper refactor at a later time, which would entail rewriting all service method definitions to access the token from the context as opposed to an argument, and rewriting all authorization/permission checks to accept/work with a `domain.Identity` as opposed to a raw token.

Currently, __only the event publisher makes use of the `domain.Identity` value from the context__. The rest of the code still follows the existing pattern where the token is extracted by a `go-kit` `DecodeRequestFunc` and then forwarded to the service methods (as an argument), after which they eventually perform `auth.Identify` calls themselves. As such, the `IdentityMiddleware` does not abort the request on error (e.g. missing auth token or a failure during `auth.Identify`), it simply logs the error. This is something that would be changed in the future.

---

> After inspecting the codebase for calls to `auth.Identify` I discovered some _existing_ code-paths that make duplicate calls to `auth.Identify`. For example, the `canUserAccessGroup` helper in `things`:
> https://github.com/MainfluxLabs/mainflux/blob/6cddb5d118f6a6b859f1508e8a2907c08cb89e5d/things/groups.go#L317-L336
> ... it first makes a call to `auth.Identify` to obtain the User ID, but then later potentially a call to `isAdmin` - which calls `auth.Authorize`, which internally calls `Identify()` again, because the `AuthzReq` includes a `token` opposed to a `domain.Identity`.
> So perhaps this is something that could be improved as well. My initial hunch is that all authz-related code should accept an _identity_ as opposed to a token.

Depends on #1235

### DIFF
--- a/auth/api/http/backup/endpoint_test.go
+++ b/auth/api/http/backup/endpoint_test.go
@@ -102,7 +102,7 @@ func newService() auth.Service {
 
 func newServer(svc auth.Service) *httptest.Server {
 	logger := logger.NewMock()
-	mux := httpapi.MakeHandler(svc, mocktracer.New(), logger)
+	mux := httpapi.MakeHandler(svc, thmocks.NewAuthService("", nil, nil), mocktracer.New(), logger)
 	return httptest.NewServer(mux)
 }
 
--- a/auth/api/http/backup/transport.go
+++ b/auth/api/http/backup/transport.go
@@ -8,8 +8,10 @@ import (
 	"github.com/MainfluxLabs/mainflux/auth"
 	"github.com/MainfluxLabs/mainflux/logger"
 	"github.com/MainfluxLabs/mainflux/pkg/apiutil"
+	"github.com/MainfluxLabs/mainflux/pkg/authn"
 	"github.com/MainfluxLabs/mainflux/pkg/errors"
 	"github.com/MainfluxLabs/mainflux/pkg/uuid"
+	"github.com/go-kit/kit/endpoint"
 	kitot "github.com/go-kit/kit/tracing/opentracing"
 	kithttp "github.com/go-kit/kit/transport/http"
 	"github.com/go-zoo/bone"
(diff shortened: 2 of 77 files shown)


manuio (approved):
LGTM

majabirmancevic (approved):

PR #1237, opened 2026-06-03
MF-1232 - attach acting user identity to redis events

### LINKED ISSUE #1232: Accompany platform events with actor identity information

bool3max:
- Un-exported `pkg/events.RedisEvent` and its methods
- Made slight improvements to the structure of event-related types
    - `EventAction` types now define the operation string in a separate `Operation()` method, meaning that they no longer have to encode in manually in `Encode()`. 
    - The base `Event` type is now a concrete `struct` and features an `.Encode()` method which takes care of encoding all of its parameters into the final `redisEvent`
    - The base `Event` type now additionally contains a `domain.Identity` field containing the identity of the user who initiated the event. The identity is attached in the event publisher's `.Publish()` method, which extracts it from the passed `context`, and attaches it to the event

Depends on #1236. Resolves #1232.

### DIFF
--- a/pkg/events/redis.go
+++ b/pkg/events/redis.go
@@ -31,8 +31,8 @@ const (
 	OrgCreate = orgPrefix + "create"
 	OrgRemove = orgPrefix + "remove"
 
-	actorIdentityUserID = "_actor_user_id"
-	actorIdentityEmail  = "_actor_user_email"
+	jwtIdentityUserID    = "jwt_identity_user_id"
+	jwtIdentityUserEmail = "jwt_identity_user_email"
 
 	ThingsStream = mainfluxPrefix + "things"
 	AuthStream   = mainfluxPrefix + "auth"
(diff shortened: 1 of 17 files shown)


pkg/events/redis.go
```diff
@@ -29,21 +31,38 @@ const (
...
+// operation returns the event's operation name, or an empty string if missing.
+func (e redisEvent) operation() string {
+	return e.field("operation", "")
+}
+
+// jwtUserIdentity returns the identity of the user that initiated the event.
+func (e redisEvent) jwtUserIdentity() domain.Identity {
+	var identity domain.Identity
+
+	if jwtUserID, ok := e[jwtIdentityUserID]; ok {
```
manuio: Can you cast here?

pkg/events/redis.go
```diff
@@ -29,21 +31,38 @@ const (
...
+		identity.ID, _ = jwtUserID.(string)
```
manuio: Don't ignore errors

manuio (changes requested):

pkg/events/redis.go
bool3max: I rewrote the method to do the map key accesses and type assertions together. We ignore the error on the type assertion because on missing identity we just return an empty `domain.Identity`.

manuio (approved):
LGTM

majabirmancevic (approved):

manuio (approved):
LGTM

Issue #1164, opened 2026-03-19
Rules: Alarm input type


manuio:
## Context

When a rule raises an alarm, there is currently no way to react to it within the rules engine. `Input.Type = "alarm"` enables rule chaining and severity-based response pipelines — e.g. a critical alarm triggers an SMS notification while a warning only sends an email.

The alarm subject pattern `alarms.<originType>.<originID>` is already produced by the alarms service. This issue wires the rules engine to consume it.

## Changes

**rules/rules.go**
- Add `AlarmInputConfig` struct with `Severity string` to filter by alarm severity

**rules/service.go**
- Subscribe to `alarms.>` when any rule with `Input.Type = "alarm"` exists in the group
- Extend `Consume()` to handle alarm subjects: parse `alarms.<originType>.<originID>` and route to matching rules
- In `toInputEvent()`: map `Alarm` fields (`ThingID`, `RuleID`, `Payload`, `Created`) to `map[string]any` for condition evaluation

**rules/postgres/init.go**
- No schema changes — `input` JSONB column covers this

## Acceptance Criteria

- A rule with `Input.Type = "alarm"` is evaluated when an alarm is published to `alarms.>`
- `AlarmInputConfig.Severity` filters which alarms trigger the rule; empty means all severities
- Condition evaluation works against alarm fields (`thing_id`, `rule_id`, `payload.*`, `created`)

Issue #1163, opened 2026-03-19
Rules: Command input type


manuio:
# Context
Depends on typed Input model and dynamic subscriptions issues. CommandInput is already defined but not wired.

# Changes
rules/subscriptions.go — add subject mapping for CommandInput:

CommandInput{ThingID} → commands.<thingID>
CommandInput{Type}    → commands.<thingID>.<type>
rules/service.go — add command case to toInputEvent normalizer, mapping command fields to map[string]any for condition evaluation.

PR #1240, opened 2026-06-05
MF-1161 - Introduce Input Config with subtopic filtering for rules

### LINKED ISSUE #1161: Rules: filter message input by subtopic
## Context

Once the typed Input model lands, `MessageInput` carries a `ThingID` but has no way to filter by subtopic. A thing may publish on multiple subtopics (`sensors/temperature`, `status/heartbeat`, `location/gps`) and a rule should only evaluate against the relevant one, not every message the device publishes.

`msg.Subtopic` already exists on `protomfx.Message` but is never checked during rule evaluation.

## Changes

**rules/rules.go**
- Add `Subtopic string` to `MessageInput` struct

**rules/service.go**
- In `Consume()`: skip rule evaluation if `rule.Input.MessageInput.Subtopic` is non-empty and does not match `msg.Subtopic`

## Acceptance Criteria

- A rule with `MessageInput.Subtopic = "sensors/temperature"` is not evaluated for messages published on `status/heartbeat`
- A rule with empty `Subtopic` evaluates against all subtopics (existing behaviour preserved)

majabirmancevic:
Resolves #1161

Adds a per-input-type config field (`input_config JSONB`) to rules, introducing `subtopic` as the first config option, used to filter message-type rules by subtopic.

Each input type (#1161 message, #1163 command, #1164 alarm) has its own set of config fields. Rather than adding a new flat column per type per issue, we introduce a single `input_config JSONB` column that stores type-specific config as a map. This keeps the schema stable, so future input type config fields are just new keys in the blob, no migration needed.

Regarding the `Subtopic` field in `Input.Config`, subtopic validation is omitted intentionally; the UI form will guide users on the correct **`dot-separated format`**.

### DIFF
--- a/rules/postgres/rules.go
+++ b/rules/postgres/rules.go
@@ -36,8 +36,8 @@ func (rr ruleRepository) Save(ctx context.Context, rls ...rules.Rule) ([]rules.R
 	}
 	defer tx.Rollback()
 
-	rq := `INSERT INTO rules (id, group_id, name, description, input_type, conditions, operator, actions)
-		VALUES (:id, :group_id, :name, :description, :input_type, :conditions, :operator, :actions);`
+	rq := `INSERT INTO rules (id, group_id, name, description, input_type, input_config, conditions, operator, actions)
+		VALUES (:id, :group_id, :name, :description, :input_type, :input_config, :conditions, :operator, :actions);`
 
 	for _, rule := range rls {
 		dbr, err := toDBRule(rule)
@@ -88,7 +88,7 @@ func (rr ruleRepository) RetrieveByGroup(ctx context.Context, groupID string, pm
 	}
 	whereClause := dbutil.BuildWhereClause(gq, nq, itq)
 
-	q := fmt.Sprintf(`SELECT id, group_id, name, description, input_type, conditions, operator, actions
+	q := fmt.Sprintf(`SELECT id, group_id, name, description, input_type, input_config, conditions, operator, actions
 		FROM rules %s
 		ORDER BY %s %s %s;`, whereClause, oq, dq, olq)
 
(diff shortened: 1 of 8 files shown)


rules/rules.go
```diff
@@ -13,14 +13,25 @@ import (
...
+	InputTypeCommand = "command"
 )
 
+type InputConfig map[string]any
+
+func (c InputConfig) Subtopic() string {
+	if c == nil {
+		return ""
+	}
+	s, _ := c["subtopic"].(string)
```
manuio, line 24: Don't ignore the error

rules/postgres/rules.go
```diff
@@ -363,12 +363,22 @@ type dbRule struct {
...
 	Conditions  []byte `db:"conditions"`
 	Operator    string `db:"operator"`
 	Actions     []byte `db:"actions"`
 }
 
 func toDBRule(r rules.Rule) (dbRule, error) {
+	inputConfig := []byte("{}")
+	if len(r.Input.Config) > 0 {
+		b, err := json.Marshal(r.Input.Config)
+		if err != nil {
```
manuio, line 376: Is there a way to avoid if inside if?

manuio (changes requested):

rules/rules.go
majabirmancevic: actually, it's a type assertion bool, so if `c["subtopic"]` is missing or not a string, `s` becomes "" and `ok` would be false;
so `ok` check is not necessary?

rules/postgres/rules.go
majabirmancevic: this is how it was done for `Metadata` fields in the things service, but an alternative could be:
```
inputConfig, err := json.Marshal(r.Input.Config)
if err != nil {
    return dbRule{}, errors.Wrap(dbutil.ErrMalformedEntity, err)
}
if r.Input.Config == nil {
    inputConfig = []byte("{}")
}
```
but then json.Marshal always runs

rules/rules.go
manuio: In that case we could even do `s, _ := c["subtopic"].(string)`

majabirmancevic: yes, as it is now

manuio: Yes but there is no need to check `if c == nil`

rules/postgres/rules.go
manuio: ok, let's keep it as it is

rules/rules.go
majabirmancevic: right! it's fixed

manuio (approved):
LGTM

duolok (approved):
LGTM

PR #1249, opened 2026-06-12
MF-1144 - add audit log service

bool3max:
The new `audit` services consumes all platform events, preserves them in a database, and exposes an HTTP API that allows users to query them.

---

* The `Event` type has been extended with additional properties:
    * `OccurredAt` - the time at which the event was published
    * `OrgID` (optional) - the organization within which the event occurred
    * `GroupID` (optional) - the group within which the event occurred 

## API endpoints

See the `audit.yml` OpenAPI schema for details.

### Query per-Org events
```
GET /orgs/<id>/events
```

Allows users with the role of "admin" (or higher) within an organization to query its events. Supports filtering by: specific event operation actor user email/id, time (to/from), as well as any event-specific data.

### Query per-Group events
```
GET /groups/<id>/events
```

Allows users with the role of "admin" (or higher) within a group to query its events.

### Query platform-wide events

```
GET /events
```

Allows platform administrators to query _all_ events, irrespective of the organization they occur in. This is the only way to query events that do not belong to any specific organization (for example, the `org_created` event).


api/openapi/audit.yml
```diff
@@ -0,0 +1,294 @@
...
+        - $ref: "#/components/parameters/From"
+        - $ref: "#/components/parameters/To"
+      responses:
+        '200':
+          $ref: "#/components/responses/ListEventsRes"
+        '400':
+          description: Failed due to malformed query parameters or missing organization ID.
+        '401':
+          description: Missing or invalid access token provided.
+        '403':
```
majabirmancevic, line 68: Is there any case of error 404?

audit/postgres/events.go
```diff
@@ -0,0 +1,251 @@
...
+func (r *eventRepository) RetrieveEvents(ctx context.Context, pm audit.PageMetadata) (audit.EventsPage, error) {
+	return r.retrieve(ctx, "", "", pm)
+}
+
+func (r *eventRepository) RetrieveEventsByOrg(ctx context.Context, orgID string, pm audit.PageMetadata) (audit.EventsPage, error) {
+	return r.retrieve(ctx, orgID, "", pm)
+}
+
+func (r *eventRepository) RetrieveEventsByGroup(ctx context.Context, groupID string, pm audit.PageMetadata) (audit.EventsPage, error) {
+	return r.retrieve(ctx, "", groupID, pm)
```
majabirmancevic: I think we could pass the `query` as a param to the helper, instead of sending the `groupID`/`orgID`
please check other services for this helper

audit/postgres/events.go
```diff
@@ -0,0 +1,251 @@
...
+		return "", nil, err
+	}
+	return "data @> :data", b, nil
+}
+
+type dbEvent struct {
+	ID             string         `db:"id"`
+	OccurredAt     time.Time      `db:"occurred_at"`
+	Operation      string         `db:"operation"`
+	ActorUserID    sql.NullString `db:"actor_user_id"`
```
majabirmancevic: or `ActorID`/`ActorEmail`

audit/service.go
```diff
@@ -0,0 +1,167 @@
...
+
+	// RetrieveEventsByOrg retrieves events belonging to a specific organization from the database
+	RetrieveEventsByOrg(ctx context.Context, orgID string, pm PageMetadata) (EventsPage, error)
+
+	// RetrieveEventsByGroup retrieves events belonging to a specific group from the database
+	RetrieveEventsByGroup(ctx context.Context, groupID string, pm PageMetadata) (EventsPage, error)
+}
+
+type Service interface {
+	// RecordEvent persists a single event to the database.
```
majabirmancevic: let's put a slightly different comment from `SaveEvent`

auth/redis/streams.go
```diff
@@ -28,7 +28,10 @@ func (es eventStore) CreateOrg(ctx context.Context, token string, org auth.Org)
 		return sorg, err
 	}
 
-	es.pub.Publish(ctx, events.OrgCreated{ID: sorg.ID})
+	es.pub.Publish(ctx, events.Event{
+		Action: events.OrgCreated{ID: sorg.ID},
+		OrgID:  sorg.ID,
```
majabirmancevic, line 33: why we need `orgID` here as well?

things/redis/events/streams.go
```diff
@@ -28,55 +28,107 @@ func (es eventStore) CreateThings(ctx context.Context, token, profileID string,
...
 }
 
 func (es eventStore) RemoveThings(ctx context.Context, token string, ids ...string) error {
 	for _, id := range ids {
+		groupID, err := es.Service.GetGroupIDByThing(ctx, id)
+		if err != nil {
+			return err
+		}
+
+		group, err := es.Service.ViewGroup(ctx, token, groupID)
```
majabirmancevic, line 118: it would be good if we also get `orgID` by `groupID` from redis here (#later)

auth/redis/streams.go
bool3max: Good question - if we do populate the OrgID field here, then org admins will be able to see the event (and thus when the org was created, I guess), otherwise not. I suppose it doesn't really matter, we don't need to set the field here.

pkg/events/redis.go
```diff
@@ -31,9 +34,16 @@ const (
...
 	OrgRemove = orgPrefix + "remove"
 
+	// General redis event map fields
 	jwtIdentityUserID    = "jwt_identity_user_id"
 	jwtIdentityUserEmail = "jwt_identity_user_email"
 
+	occurredAt = "occurred_at"
+
+	evOrgID   = "e_org_id"
+	evGroupID = "e_group_id"
```
manuio: The consts and the values are not consistents. Rename to `evtOrgID = evt_org_id`

manuio (changes requested):

things/redis/events/streams.go
manuio: Reference an issue here

duolok (dismissed):
LGTM

audit/service.go
```diff
@@ -71,7 +71,7 @@ type EventRepository interface {
 }
 
 type Service interface {
-	// RecordEvent persists a single event to the database.
+	// RecordEvent records a single event to the database.
```
majabirmancevic: no need to mention "to the database" for service method

things/redis/events/streams.go
majabirmancevic: related issue: https://github.com/MainfluxLabs/mainflux/issues/1253

audit/api/http/responses.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+type eventRes struct {
+	ID         string          `json:"id"`
+	OccurredAt time.Time       `json:"occurred_at"`
+	Operation  string          `json:"operation"`
+	Actor      domain.Identity `json:"actor"`
+	OrgID      string          `json:"org_id,omitempty"`
+	GroupID    string          `json:"group_id,omitempty"`
+	Data       map[string]any  `json:"data,omitempty"`
```
majabirmancevic: ActionData

audit/api/http/transport.go
```diff
@@ -0,0 +1,194 @@
...
+	kitot "github.com/go-kit/kit/tracing/opentracing"
+	kithttp "github.com/go-kit/kit/transport/http"
+	"github.com/go-zoo/bone"
+	"github.com/opentracing/opentracing-go"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+const (
+	operationKey = "operation"
+	dataKey      = "data"
```
majabirmancevic: action_data or not?

audit/service.go
```diff
@@ -0,0 +1,167 @@
...
+
+type PageMetadata struct {
+	Total     uint64         `json:"total,omitempty"`
+	Offset    uint64         `json:"offset,omitempty"`
+	Limit     uint64         `json:"limit,omitempty"`
+	Order     string         `json:"order,omitempty"`
+	Dir       string         `json:"dir,omitempty"`
+	Email     string         `json:"email,omitempty"`
+	Operation string         `json:"operation,omitempty"`
+	Data      map[string]any `json:"data,omitempty"`
```
majabirmancevic: here?

api/openapi/audit.yml
```diff
@@ -0,0 +1,330 @@
...
+      tags:
+        - audit
+      parameters:
+        - $ref: "#/components/parameters/Offset"
+        - $ref: "#/components/parameters/Limit"
+        - $ref: "#/components/parameters/Order"
+        - $ref: "#/components/parameters/Dir"
+        - $ref: "#/components/parameters/Email"
+        - $ref: "#/components/parameters/Operation"
+        - $ref: "#/components/parameters/Data"
```
majabirmancevic: ActionData

api/openapi/audit.yml
```diff
@@ -0,0 +1,330 @@
...
+          $ref: "#/components/schemas/Actor"
+        org_id:
+          type: string
+          format: uuid
+          description: Organization the event belongs to. Absent for events not tied to any organization.
+        group_id:
+          type: string
+          format: uuid
+          description: Group the event belongs to. Absent for events not tied to any group.
+        data:
```
majabirmancevic: and here (check the rest of the file)

audit/postgres/events.go
```diff
@@ -0,0 +1,297 @@
...
+
+func operationQuery(op string) (string, string) {
+	if op == "" {
+		return "", ""
+	}
+	return "operation = :operation", op
+}
+
+func orgQuery(orgID string) (string, string) {
+	if orgID == "" {
```
anna02272, line 197: These functions are only called from RetrieveEventsByOrg and RetrieveEventsByGroup, which always receive ID (validated in the request layer). So I guess we don't need this guard?

manuio: Good question

manuio: Actually, it's good. Let's keeop it as it is.

things/redis/events/streams.go
```diff
@@ -28,55 +28,107 @@ func (es eventStore) CreateThings(ctx context.Context, token, profileID string,
...
-		ID:        thing.ID,
-		ProfileID: thing.ProfileID,
-		GroupID:   thing.GroupID,
+	es.pub.Publish(ctx, events.Event{
+		Action: events.ThingGroupAndProfileUpdated{
+			ID:        thing.ID,
+			ProfileID: thing.ProfileID,
+			GroupID:   thing.GroupID,
+		},
+		GroupID: prevGroupID,
```
majabirmancevic, line 104: there shouldn't be any confusion about `prevGroupID` and the new `thing.GroupID` (when we access them)?

bool3max: Yeah I was trying to think of what makes the most sense for this event. Once the thing's profile and/or group are changed, and the event is emitted:


1. Should the ProfileID and GroupID fields of the actual event be those of the profile and group that the thing was moved to, or those of the profile and group it was moved from? I think quite obviously the former, i.e. the event should tell us where the thing was moved. That's the way the code currently is..
2. The less obvious choice, is which group the event should belong to..
    a. If the thing's group wasn't changed (i.e. only the profile was), it's pretty clear, the event is emitted in the thing's belonging group, which, again, doesn't change.
    b. If the thing was moved to a different group however, to which group should the event emitted event belong? The group the thing was in before it was moved, or the destination one it was moved to? .
        i. If it's the destination one, then there really isn't a way for end users/admins to know which group it belonged to before it was moved. .
        ii. If it's the thing's previous group, then its admins can clearly see that there was once a thing in this group, and which group it was moved to. .

I think much of the confusion can be alleviated here by extending the data of the actual event to contain the IDs of both the thing's previous group and previous profile. However even then there's the question of which group the event should belong to.

auth/redis/streams.go
majabirmancevic: let's leave it as-is

majabirmancevic (approved):

anna02272 (approved):
LGTM

manuio (approved):
LGTM

PR #1251, opened 2026-06-17
NOISSUE - converters: validate CSV before async publish

manuio:


converters/api/endpoint.go
```diff
@@ -32,9 +31,8 @@ func convertCSVToSenMLEndpoint(svc converters.Service) endpoint.Endpoint {
...
 		}
 
-		if err := svc.PublishSenMLMessages(ctx, req.key.Value, req.csvLines); err != nil {
-			return nil, err
-		}
+		// Publish async to avoid blocking past the gateway timeout on large files.
+		go func() { _ = svc.PublishSenMLMessages(context.Background(), req.key.Value, req.csvLines) }()
 
 		return nil, nil
 	}
```
duolok, line 38: Could we add channels here for concurrent error handling?

manuio: channels only make sense when something reads from them

majabirmancevic (approved):

duolok (approved):

PR #1255, opened 2026-06-18
MF-1250 - Add JSON converter endpoints, consolidate routes to ?to= query param, and extend SenML to support reader exports

### LINKED ISSUE #1250: Add JSON file input endpoints for SenML and JSON conversion
Add endpoints that accept a JSON array of objects via multipart upload and publish them as SenML or JSON messages, mirroring the existing /csv/* endpoints.

anna02272:
Resolves #1250 

- Add POST /json/convert?to=senml and POST /json/convert?to=json converter endpoints that accept a JSON file via multipart/form-data, mirroring the existing CSV endpoints

- Consolidate four path-based routes (/csv/senml, /csv/json, /json/senml, /json/json) into two routes (/csv/convert, /json/convert) with a required ?to=senml|json query parameter selecting the output format

- Extend SenML converter to accept both standard SenML field names (n, t, v, vs, vd, vb, u, s) and reader-exported field names (name, time, value, string_value, data_value, bool_value, unit, sum); reader-exported time values (nanoseconds) are automatically converted to seconds

- Strip envelope fields (protocol, publisher, subtopic, created, Created) from all converted records so they don't leak into the stored payload

- Auto-unwrap reader-exported Format A records ({"payload":{...},"created":...}) in the JSON→JSON path — both flat upload files and reader exports produce identical results

- Rename PublishSenMLMessages/PublishJSONMessages to PublishSenMLMessagesFromCSV/PublishJSONMessagesFromCSV for consistency with the new JSON variants

### DIFF
--- a/converters/api/transport.go
+++ b/converters/api/transport.go
@@ -61,12 +61,6 @@ func MakeHandler(svc adapter.Service, ac domain.AuthClient, tracer opentracing.T
 	return r
 }
 
-type SenmlRec struct {
-	Name  string `json:"n"`
-	Time  string `json:"t"`
-	Value string `json:"v"`
-}
-
 func decodeConvertCSVFile(_ context.Context, r *http.Request) (any, error) {
 	if !strings.Contains(r.Header.Get("Content-Type"), multiPartContentType) {
 		return nil, apiutil.ErrUnsupportedContentType
@@ -81,6 +75,7 @@ func decodeConvertCSVFile(_ context.Context, r *http.Request) (any, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer file.Close()
 
 	csvLines, readErr := csv.NewReader(file).ReadAll()
 	if readErr != nil {
@@ -110,6 +105,7 @@ func decodeConvertJSONFile(_ context.Context, r *http.Request) (any, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer file.Close()
 
 	data, err := io.ReadAll(file)
 	if err != nil {
--- a/converters/api/transport.go
+++ b/converters/api/transport.go
@@ -6,6 +6,8 @@ package api
 import (
 	"context"
 	"encoding/csv"
+	"encoding/json"
+	"io"
 	"net/http"
 	"strings"
 
@@ -53,6 +55,20 @@ func MakeHandler(svc adapter.Service, ac domain.AuthClient, tracer opentracing.T
 		opts...,
 	))
 
+	r.Post("/json/senml", kithttp.NewServer(
+		kitot.TraceServer(tracer, "convert_json_to_senml")(convertJSONToSenMLEndpoint(svc)),
+		decodeConvertJSONFile,
+		encodeResponse,
+		opts...,
+	))
+
+	r.Post("/json/json", kithttp.NewServer(
+		kitot.TraceServer(tracer, "convert_json_to_json")(convertJSONToJSONEndpoint(svc)),
+		decodeConvertJSONFile,
+		encodeResponse,
+		opts...,
+	))
+
 	r.GetFunc("/health", mainflux.Health("converters"))
 	r.Handle("/metrics", promhttp.Handler())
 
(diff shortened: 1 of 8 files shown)


converters/api/transport.go
```diff
@@ -53,6 +55,20 @@ func MakeHandler(svc adapter.Service, ac domain.AuthClient, tracer opentracing.T
...
 	))
 
+	r.Post("/json/senml", kithttp.NewServer(
+		kitot.TraceServer(tracer, "convert_json_to_senml")(convertJSONToSenMLEndpoint(svc)),
+		decodeConvertJSONFile,
+		encodeResponse,
+		opts...,
+	))
+
+	r.Post("/json/json", kithttp.NewServer(
```
anna02272: Not really sure about this, it's consitent with ```/csv/json```

majabirmancevic: maybe we can also consider this option with query param:
```
POST /csv/convert?to=senml
POST /csv/convert?to=json
POST /json/convert?to=senml
POST /json/convert?to=json
```
one decoder per input format (as now), the output format is read from the query param

converters/service.go
```diff
@@ -334,113 +353,84 @@ func toSenMLEntries(record map[string]any) ([]map[string]any, error) {
...
+		var ts float64
+		if timeField != "" {
+			ts, _ = inputRecord[timeField].(float64)
 		}
-
-		record := map[string]any{}
-		if outerTimestamp != 0 {
-			record["Created"] = outerTimestamp
+		if ts == 0 {
+			ts, _ = inputRecord["created"].(float64)
```
manuio: Don't ignor the error

converters/service.go
```diff
@@ -334,113 +353,84 @@ func toSenMLEntries(record map[string]any) ([]map[string]any, error) {
...
+// msg.Protocol / msg.Subtopic from envelope fields. Handles both Format A
+// (reader-exported {"payload":{...},"created":...}) and flat records.
+func parseJSONRecord(inputRecord map[string]any, timeField string, msg *protomfx.Message) map[string]any {
+	record := map[string]any{}
+
+	if pld, ok := inputRecord["payload"].(map[string]any); ok {
+		// Format A — unwrap outer envelope, prefer timeField over "created".
+		var ts float64
+		if timeField != "" {
+			ts, _ = inputRecord[timeField].(float64)
```
manuio: Don't ignore the error

converters/service.go
```diff
@@ -334,113 +353,84 @@ func toSenMLEntries(record map[string]any) ([]map[string]any, error) {
...
+	if err != nil {
+		return err
 	}
-	return nil
+	msg.Payload = data
+	_, err = as.publish(ctx, key, msg)
+	return err
 }
 
 func (as *adapterService) publish(ctx context.Context, key string, msg protomfx.Message) (m protomfx.Message, err error) {
```
manuio: Should we only return an error?

converters/service.go
```diff
@@ -334,113 +353,84 @@ func toSenMLEntries(record map[string]any) ([]map[string]any, error) {
...
+
+// publishMsgs marshals a batch of records, sets msg.Payload, and publishes.
+func (as *adapterService) publishMsgs(ctx context.Context, key string, msgs []map[string]any, msg protomfx.Message) error {
+	data, err := json.Marshal(msgs)
+	if err != nil {
+		return err
 	}
-	return nil
+	msg.Payload = data
+	_, err = as.publish(ctx, key, msg)
```
manuio: Should we fix publish to don't have to ignore the `Message`?

manuio (changes requested):

converters/service.go
```diff
@@ -158,35 +192,272 @@ func (as *adapterService) PublishJSONMessages(ctx context.Context, key string, c
...
+	if pld, ok := inputRecord["payload"].(map[string]any); ok {
+		// Format A — unwrap outer envelope, prefer timeField over "created".
+		var ts float64
+		if timeField != "" {
+			if t, ok := inputRecord[timeField].(float64); ok {
+				ts = t
+			}
+		}
+		if ts == 0 {
+			if t, ok := inputRecord["created"].(float64); ok {
```
manuio: `created` or `Created`?

manuio (changes requested):

converters/service.go
anna02272: "created" is for reading from the raw input.  JSON has {"created": <nanoseconds>...} with a lowercase key.

"Created" is for writing to the output map. This is key that ToJSONMessage() in pubsub.go looks for to set the stored timestamp.

manuio (approved):
LGTM

PR #1256, opened 2026-06-19
MF-1141 - Add Device Shadow service

### LINKED ISSUE #1141: Add Device Shadow Service for Persistent Thing State Management
Add a new shadows service that maintains a persistent record of each Thing's last known state (reported) and desired target state (desired). The service will subscribe to Thing messages using the existing Consumer interface to continuously update the reported state — including any attribute a Thing reports such as sensor readings or last known location — expose REST endpoints to set desired state per Thing, and publish the full desired state back to the Thing via the existing Publisher interface — ensuring pending commands are delivered automatically upon reconnection regardless of the transport protocol used.

The desired state represents what an operator wants a Thing to be doing. When a Thing reconnects, it receives the full desired payload and is responsible for applying any changes against its own current state. For example:

```
{
  "thing_id": "3e4f-...",
  "reported": {
    "temperature": 19,
    "fan": "off",
    "location": {
      "lat": 40.7128,
      "lon": -74.0060
    }
  },
  "desired": {
    "temperature": 22,
    "fan": "on"
  },
  "reported_at": "2026-03-16T10:32:00Z",
  "updated_at": "2026-03-16T10:45:00Z"
}
```

majabirmancevic:
Adds a shadows service that maintains a persistent device shadow per thing: the operator's desired state and the device's last-known reported state, with a derived delta pushed to the device so pending config is delivered even across reconnects. 

**API**
`PUT /things/:id/shadows` - set desired state (returns the shadow + delta)
`GET /things/:id/shadows` - read the shadow
`DELETE /things/:id/shadows` - remove the shadow

**How it works**
**Desired** ➡ device: on `PUT`, the service computes `delta = desired − reported` and publishes it as a command to` things.<id>.commands.shadow`.
**Reported** ⬅ device: consumes thing telemetry off the broker and merges it into reported (JSON and SenML payloads supported).

**Reconnect**: the broker doesn't retain, so when a device next reports, any still-open delta is re-published - the device catches up without a GET.

Closes #1141 

Add:
- [x] Tests
- [x] Docs


shadows/postgres/shadows.go
```diff
@@ -0,0 +1,138 @@
...
+
+func (sr shadowRepository) Upsert(ctx context.Context, shadow shadows.Shadow) (shadows.Shadow, error) {
+	dbSh, err := toDBShadow(shadow)
+	if err != nil {
+		return shadows.Shadow{}, errors.Wrap(dbutil.ErrCreateEntity, err)
+	}
+
+	q := `INSERT INTO shadows (thing_id, desired, reported, reported_at, updated_at)
+	      VALUES (:thing_id, :desired, :reported, :reported_at, :updated_at)
+	      ON CONFLICT (thing_id) DO UPDATE SET
```
manuio, line 37: Is this to support the first shadow saving?

shadows/postgres/shadows.go
```diff
@@ -0,0 +1,138 @@
...
+	          reported_at = EXCLUDED.reported_at,
+	          updated_at  = EXCLUDED.updated_at;`
+
+	if _, err := sr.db.NamedExecContext(ctx, q, dbSh); err != nil {
+		pgErr, ok := err.(*pgconn.PgError)
+		if ok {
+			switch pgErr.Code {
+			case pgerrcode.InvalidTextRepresentation:
+				return shadows.Shadow{}, errors.Wrap(dbutil.ErrMalformedEntity, err)
+			case pgerrcode.UniqueViolation:
```
manuio: `pgerrcode.UniqueViolation` is probably dead code

shadows/postgres/shadows.go
majabirmancevic: yes, plus every subsequent update - it's a create-or-update;
the shadow row can first appear from either an operator `desired` update or a device report,  so the atomic upsert avoids a create/update race

shadows/api/http/responses.go
```diff
@@ -0,0 +1,68 @@
...
+	_ apiutil.Response = (*removeRes)(nil)
+)
+
+type stateRes struct {
+	Desired  shadows.State `json:"desired"`
+	Reported shadows.State `json:"reported"`
+	Delta    shadows.State `json:"delta,omitempty"`
+}
+
+type shadowResponse struct {
```
manuio: `shadowRes`

shadows/api/http/requests.go
```diff
@@ -0,0 +1,48 @@
...
+	if req.token == "" {
+		return apiutil.ErrBearerToken
+	}
+
+	if req.thingID == "" {
+		return apiutil.ErrMissingThingID
+	}
+
+	if len(req.Desired) == 0 {
+		return apiutil.ErrEmptyList
```
manuio: Shouldn't we use another error here?

manuio (changes requested):

shadows/postgres/init.go
```diff
@@ -0,0 +1,57 @@
...
+func migrateDB(db *sqlx.DB) error {
+	migrations := &migrate.MemoryMigrationSource{
+		Migrations: []*migrate.Migration{
+			{
+				Id: "shadows_1",
+				Up: []string{
+					`CREATE TABLE IF NOT EXISTS shadows (
+						thing_id    UUID PRIMARY KEY,
+						desired     JSONB  NOT NULL DEFAULT '{}',
+						reported    JSONB  NOT NULL DEFAULT '{}',
```
duolok, line 46: Should we maybe use `desired_state` and `reported_state` here?

majabirmancevic: hm, maybe, but I followed the naming from the issue example
https://github.com/MainfluxLabs/mainflux/issues/1141#issue-4083916416

shadows/service.go
```diff
@@ -0,0 +1,222 @@
...
+// writes) and re-publishes any pending delta, so a reconnecting device
+// receives missed commands.
+func (ss *shadowsService) ConsumeMessage(_ string, msg protomfx.Message) error {
+	patch, ok := decodeState(msg)
+	if !ok || len(patch) == 0 {
+		return nil
+	}
+
+	ctx := context.Background()
+	current, err := ss.shadows.RetrieveByThing(ctx, msg.Publisher)
```
majabirmancevic, line 117: `UpdateDesiredState` reads `current.Reported` then upserts the complete row (including that stale `Reported`). 
A concurrent `ConsumeMessage` that writes new reported state between those two calls gets silently overwritten. 
Fix it.

shadows/events/handler.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+// NewEventHandler returns new event store handler.
+func NewEventHandler(svc shadows.Service) events.EventHandler {
+	return &eventHandler{svc: svc}
+}
+
+func (h *eventHandler) Handle(ctx context.Context, event events.Event) error {
+	switch e := event.Action.(type) {
+	case events.ThingRemoved:
```
anna02272, line 24: Should we clean up shadows when a group is deleted ?

```case events.GroupRemove:```

File: shadows/postgres/shadows.go [hunk 8]
```diff
@@ -0,0 +1,148 @@
...
+func (sr shadowRepository) RetrieveByThing(ctx context.Context, thingID string) (shadows.Shadow, error) {
+	q := `SELECT thing_id, desired, reported, reported_at, updated_at
+	      FROM shadows WHERE thing_id = $1;`
+
+	dbSh := dbShadow{}
+	if err := sr.db.QueryRowxContext(ctx, q, thingID).StructScan(&dbSh); err != nil {
+		pgErr, ok := err.(*pgconn.PgError)
+		// A thing's shadow always exists conceptually: a missing row
+		// reads back as an empty shadow rather than an error.
+		if err == sql.ErrNoRows || (ok && pgerrcode.InvalidTextRepresentation == pgErr.Code) {
```
anna02272, line 98: This currently returns ` "desired": null and "reported": null `instead of `{}` when a thing has never had a shadow.

I think it would be better to return `{}` instead of `null` for empty desired and reported states.

We can fix this by explicitly setting `Desired: shadows.State{} and Reported: shadows.State{}` in the return statement.

I’m testing this with the UI, and I think displaying an empty object looks better than `null` in the code editor. I can handle this on the UI side as well, but I think it makes more sense to fix it here.

majabirmancevic: sure, let's fix that in the backend

anna02272 (approved):
LGTM

duolok (approved):
LGTM

manuio (approved):
LGTM

PR #1260, opened 2026-06-29
MF-1257 - Implement additional filters across entities

### LINKED ISSUE #1257: Add missing filters across entities
Several entities currently support only a subset of the filters that would be useful for users.

`Things`
Currently filterable: name, metadata
Needs backend:  type

`Profiles`
Currently filterable: name, metadata
Needs backend: content_type

`Org Memberships`
Currently filterable: email
Needs backend: role

`Group Memberships`
Currently filterable: email
Needs backend: role

`Alarms`
Currently filterable: level, status
Needs backend: time range, protocol, subtopic

`Script Runs`
Currently filterable: none
Needs backend: name, status, time range

`Webhooks`
Currently filterable: name
Needs backend: url

`Downlinks`
Currently filterable: name
Needs backend: method, url, frequency

`Notifiers`
Currently filterable: name, metadata
Needs backend: contact

`Files`
Currently filterable: name, class, format,
Needs backend:  time 

`Certs`
Currently filterable: none
Needs backend: serial, expires_at

`Modbus`
Currently filterable: name
Needs backend: ip_address, port, slave_id, frequency

`Commands` 
Currently filterable: none
Needs backend: topic, thing_id, created_at

anna02272:
Resolves #1257 

This PR adds support for previously missing filters across multiple entities, including Things (type), Profiles (content_type), Org and Group Memberships (role), Alarms (time range, protocol, subtopic), Script Runs (status, time range), Webhooks (url), Downlinks (method, url, frequency), Notifiers (contact), Certs (serial, expires_at), Modbus (ip_address, port, slave_id, frequency), and Commands (topic).


certs/postgres/certs.go
```diff
@@ -355,6 +361,17 @@ func (cr certsRepository) MarkDownloaded(ctx context.Context, serial string) err
...
+	var parts []string
+	if !from.IsZero() {
+		parts = append(parts, "expires_at >= :from")
+	}
+	if !to.IsZero() {
+		parts = append(parts, "expires_at < :to")
+	}
+	return strings.Join(parts, " AND ")
+}
+
```
duolok, line 374: Should we add `from > to` check here or to the request layer? It seems like we're missing that

manuio: I guess that nothing would be returned but we can add the check in case

anna02272: Yes, nothing is returned. I initially wanted to add it, but then I checked how it's handled for messages, and since we don't return it there either, I decided to keep it consistent. We can add it, though.

manuio: I'd say that it's out off scope for this PR. However, we can create an issue and fix it for any from/to filter.

duolok (approved):
LGTM

auth/memberships.go
```diff
@@ -138,44 +138,52 @@ func (svc service) ListOrgMemberships(ctx context.Context, token string, orgID s
...
 
-	if len(memberships) == 0 {
-		return OrgMembershipsPage{
-			OrgMemberships: []OrgMembership{},
-			Total:          0,
-		}, nil
+	if pm.Role != "" {
+		var filtered []OrgMembership
+		for _, m := range memberships {
+			if m.Role == pm.Role {
```
majabirmancevic, line 144: not sure if this is a good place to filter, is it possible to pass it directly to the SQL query (like `orgID`)?

majabirmancevic: btw, should we use some other method instead of `BackupByOrg` to retrieve memberships?

manuio: Right, there is an issue to don't use BackupByOrg: https://github.com/MainfluxLabs/mainflux/issues/1261

majabirmancevic (approved):

PR #1263, opened 2026-06-30
MF-1262 - Make the to timestamp filter exclusive for messages

### LINKED ISSUE #1262: Make the "to" timestamp filter exclusive for messages
The current message filtering treats both `from` and `to` as inclusive. Update the `to` filter to be exclusive (created < :to) while keeping `from` inclusive.

anna02272:
Resolves #1262

### DIFF
--- a/readers/postgres/json_test.go
+++ b/readers/postgres/json_test.go
@@ -18,6 +18,11 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const (
+	fromIndex = 20
+	toIndex   = 50
+)
+
 func TestListJSONMessages(t *testing.T) {
 	reader := preader.NewJSONRepository(db)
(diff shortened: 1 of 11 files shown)


readers/postgres/json_test.go
```diff
@@ -332,7 +332,7 @@ func TestDeleteJSONMessages(t *testing.T) {
 			pageMeta: readers.JSONPageMetadata{
 				Publisher: id1,
 				From:      int64(created + 20),
-				To:        int64(created + 50),
+				To:        int64(created + 51),
```
manuio: Should we use something like ` int64(created + toIndex + 1),`?

manuio (approved):
LGTM

duolok (approved):
LGTM

PR #1269, opened 2026-07-03
MF-1268 - Fix converting CSV files with BOM

### LINKED ISSUE #1268: converters: CSV upload fails when file has UTF-8 BOM
CSV files exported from Excel or Windows tools include a hidden UTF-8 BOM (0xEF 0xBB 0xBF) prepended to the first column header. This causes time_field column lookup to fail silently (time_field_idx=-1), so all records are stored with created = now instead of the per-row timestamp.

manuio:
Resolves #1268

### DIFF
--- a/converters/api/transport.go
+++ b/converters/api/transport.go
@@ -4,6 +4,7 @@
 package api
 
 import (
+	"bytes"
 	"context"
 	"encoding/csv"
 	"encoding/json"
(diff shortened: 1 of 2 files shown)


duolok (approved):
LGTM

anna02272 (approved):

PR #1272, opened 2026-07-07
MF-1142 - Upgrade the internal message bus to NATS JetStream 

### LINKED ISSUE #1142: Upgrade Internal Message Bus to NATS JetStream for Durable Message Delivery
Currently the platform uses plain NATS pub/sub with fire-and-forget semantics in pkg/messaging/nats, meaning messages are dropped if any consumer service is temporarily unavailable. Upgrade the internal message bus to NATS JetStream to introduce persistent, durable streams with at-least-once delivery guarantees across all services that depend on the message bus — consumers, rules, alarms, webhooks and notifiers.

duolok:
Now, messages won't be dropped even if consumer service goes down. 

Resolves #1142


pkg/messaging/nats/connection.go
```diff
@@ -0,0 +1,85 @@
+// Copyright (c) Mainflux
+// SPDX-License-Identifier: Apache-2.0
+
```
majabirmancevic, line 3: maybe we could use a noun for the file name? (`connect.go`->`client.go`, for example)

duolok: That entire file is used as a connection helper, so maybe `connection.go` would be a suitable name?

majabirmancevic: aha, sure!

pkg/messaging/nats/connect.go
```diff
@@ -0,0 +1,85 @@
...
+	if err := ensureStream(js); err != nil {
+		conn.Close()
+		return nil, nil, err
+	}
+
+	return conn, js, nil
+}
+
+func ensureStream(js broker.JetStreamContext) error {
+	switch _, err := js.StreamInfo(streamName); {
```
majabirmancevic: do we really need switch-case here, or can it be simplified?

pkg/messaging/nats/connection.go
```diff
@@ -0,0 +1,85 @@
...
+var durableReplacer = strings.NewReplacer(
+	".", "_",
+	"*", "all",
+	">", "gt",
+	" ", "_",
+	"/", "_",
+	"\\", "_",
+)
+
+func sanitizeDurable(s string) string {
```
majabirmancevic: is this helper necessary? (there is one call?)

pkg/messaging/nats/connection.go
```diff
@@ -0,0 +1,84 @@
...
+		return err
+	}
+
+	_, err = js.AddStream(&broker.StreamConfig{
+		Name:     streamName,
+		Subjects: streamSubjects,
+		Storage:  broker.FileStorage,
+	})
+
+	if err != nil && strings.Contains(err.Error(), "already in use") {
```
manuio: Use a const for `"already in use"` or if it's possible use a nats error.

pkg/messaging/nats/connection.go
```diff
@@ -0,0 +1,84 @@
...
+		return nil
+	}
+	return err
+}
+
+func durableName(queue, id, topic string) string {
```
manuio, line 69: What is this method for?

manuio (changes requested):

pkg/messaging/nats/connection.go
duolok: JetStreams forbids usage of some characters when creating a name for a stream, so this method replaces those characters.

https://docs.nats.io/running-a-nats-service/nats_admin/jetstream_admin/naming

majabirmancevic (approved):

manuio (approved):
LGTM

PR #1274, opened 2026-07-08
NOISSUE - Remove unused command input type

majabirmancevic:
The `rules` service defined `command` as a valid input type alongside `message` and `alarm`, but no subscription or consumer was ever wired for it - a rule created with `input_type: "command"` was accepted but would never fire. 
The command input was an early idea (#1163) that has since been dropped, so this PR removes the unused definitions.

### DIFF
--- a/api/openapi/rules.yml
+++ b/api/openapi/rules.yml
@@ -446,7 +446,7 @@ components:
       properties:
         type:
           type: string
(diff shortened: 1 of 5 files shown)


manuio (approved):
LGTM

duolok (approved):
LGTM

PR #1279, opened 2026-07-16
MF-1143 - Publishing Will message and handling duplicate messages

### LINKED ISSUE #1143: Honor MQTT QoS Levels in the MQTT Adapter
Currently the MQTT adapter via mProxy parses but completely ignores QoS 0/1/2 flags, retained messages, persistent sessions, and Last Will messages sent by devices, effectively downgrading all connections to QoS 0. Implement proper QoS handling in the MQTT adapter so that device-requested delivery guarantees are respected at the transport layer, including persistent session management and retained message storage.

duolok:
Originally, I wanted to cover all the missing pieces described in the issue. After thorough examination, I realized that only fixing double writes with DUP and handling Last Will is what we were missing. This PR is adding those two features to the codebase.

| Issue claim | What I found | Status |
|---|---|---|
| QoS 0/1/2 flags | Mosquitto already handles these | not broken |
| Retained messages ignored | Mosquitto stores and replays them | not broken |
| Persistent sessions ignored | works — Mosquitto respects `clean_session=false` | not broken |
| Last Will ignored | true, and it lets a thing bypass command auth | fixed |
| Everything downgraded to QoS 0 | nothing downgrades anything; the forwarder uses QoS 2 | not broken |
| DUP retransmits (not in the issue) | these were getting written to the DB twice | fixed |

Resolves #1143

### DIFF
--- a/mqtt/handler.go
+++ b/mqtt/handler.go
@@ -208,7 +208,7 @@ func (h *handler) Publish(c *session.Client, topic *string, payload *[]byte) {
 	}
 
 	if err := h.publishToBus(c, *topic, *payload); err != nil {
-		h.logger.Error(err.Error())
+		h.logger.Error(fmt.Sprintf("client_id %s failed to publish to topic %s: %s", c.ID, *topic, err))
 		return
 	}
 
--- a/mqtt/handler.go
+++ b/mqtt/handler.go
@@ -207,7 +207,7 @@ func (h *handler) Publish(c *session.Client, topic *string, payload *[]byte) {
 	}
 
 	if err := h.publishToBus(c, *topic, *payload); err != nil {
-		h.logger.Error(fmt.Sprintf("client_id %s failed to publish to topic %s", c.ID, *topic))
+		h.logger.Error(fmt.Sprintf("client_id %s failed to publish to topic %s: %s", c.ID, *topic, err))
 		return
 	}
 
@@ -222,7 +222,7 @@ func (h *handler) publishToBus(c *session.Client, topic string, payload []byte) 
 
 	pc, err := h.things.GetPubConfigByKey(context.Background(), tk)
 	if err != nil {
-		return errors.Wrap(messaging.ErrPublishMessage, err)
+		return err
 	}
 
 	subject, subtopic, err := parseTopic(topic, pc.PublisherID)
@@ -237,18 +237,18 @@ func (h *handler) publishToBus(c *session.Client, topic string, payload []byte) 
 	}
 
 	if err := messaging.FormatMessage(pc, &msg); err != nil {
-		return errors.Wrap(messaging.ErrPublishMessage, err)
+		return err
 	}
 
 	if isCommandSubject(subject) {
 		if err := h.publishCommand(subject, msg); err != nil {
-			return errors.Wrap(messaging.ErrPublishMessage, err)
+			return err
 		}
 		return nil
 	}
 
 	if err := h.publishMessage(pc, msg); err != nil {
-		return errors.Wrap(messaging.ErrPublishMessage, err)
+		return err
 	}
 
 	return nil
--- a/mqtt/handler.go
+++ b/mqtt/handler.go
@@ -29,8 +29,8 @@ const (
 	topicSuffixCommands = "commands"
 	topicSuffixMessages = "messages"
 
-	typeCommand = "command"
-	typeMessage = "message"
+	kindMessage = "message"
+	kindCommand = "command"
 )
 
 var (
@@ -220,7 +220,7 @@ func (h *handler) Publish(c *session.Client, topic *string, payload *[]byte) {
 }
 
 // publishToBus routes a client's topic and payload onto the internal message bus,
-// returning "command" or "message" to describe what was published.
+// reporting the kind of what was published.
 func (h *handler) publishToBus(c *session.Client, topic string, payload []byte) (string, error) {
 	tk := domain.ThingKey{
 		Value: string(c.Password),
@@ -251,14 +251,14 @@ func (h *handler) publishToBus(c *session.Client, topic string, payload []byte) 
 		if err := h.publishCommand(subject, msg); err != nil {
 			return "", errors.Wrap(messaging.ErrPublishMessage, err)
 		}
-		return typeCommand, nil
+		return kindCommand, nil
 	}
 
 	if err := h.publishMessage(pc, msg); err != nil {
 		return "", errors.Wrap(messaging.ErrPublishMessage, err)
 	}
 
-	return typeMessage, nil
+	return kindMessage, nil
 }
 
 func (h *handler) publishCommand(subject string, msg protomfx.Message) error {
--- a/mqtt/handler.go
+++ b/mqtt/handler.go
@@ -383,7 +383,7 @@ func (h *handler) Disconnect(c *session.Client) {
 		return
 	}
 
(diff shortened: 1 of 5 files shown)


mqtt/handler.go
```diff
@@ -365,13 +375,30 @@ func (h *handler) Disconnect(c *session.Client) {
...
+
 	if err := h.cache.Disconnect(context.Background(), c.ID); err != nil {
 		h.logger.Error(errors.Wrap(errFailedCacheDisconnection, err).Error())
 	}
 
 	h.logger.Info(fmt.Sprintf("client_id %s disconnected", c.ID))
 }
 
+// publishWill mirrors a client's Last Will onto the internal bus.
+func (h *handler) publishWill(c *session.Client) {
```
majabirmancevic: is this helper necessary, or could we call `publishToBus` directly in `Disconnect`?

mqtt/handler.go
```diff
@@ -206,49 +210,55 @@ func (h *handler) Publish(c *session.Client, topic *string, payload *[]byte) {
 		return
 	}
 
+	kind, err := h.publishToBus(c, *topic, *payload)
```
majabirmancevic: I'd avoid returning `kind` here (since there is no special logic based on that value); instead, we could:

```
func (h *handler) Publish(c *session.Client, topic *string, payload *[]byte) {
   ...

    if err := h.publishToBus(c, *topic, *payload); err != nil {
		h.logger.Error(err.Error())
		return
	}
	h.logger.Info(fmt.Sprintf("client_id %s published to topic %s", c.ID, *topic))
}
```
since we know if it's a message or a command based on the topic?

mqtt/handler.go
```diff
@@ -206,49 +207,52 @@ func (h *handler) Publish(c *session.Client, topic *string, payload *[]byte) {
...
-		h.logger.Error(errors.Wrap(messaging.ErrPublishMessage, err).Error())
-		return
+		return errors.Wrap(messaging.ErrPublishMessage, err)
 	}
 
 	if isCommandSubject(subject) {
 		if err := h.publishCommand(subject, msg); err != nil {
-			h.logger.Error(errors.Wrap(messaging.ErrPublishMessage, err).Error())
-			return
+			return errors.Wrap(messaging.ErrPublishMessage, err)
```
majabirmancevic: we can consider introducing separate Error messages for Commands/Webhooks/.. (#later - new issue)

mqtt/handler.go
```diff
@@ -206,49 +207,52 @@ func (h *handler) Publish(c *session.Client, topic *string, payload *[]byte) {
 		return
 	}
 
+	if err := h.publishToBus(c, *topic, *payload); err != nil {
+		h.logger.Error(err.Error())
```
majabirmancevic: can we include `client_id` in the log?

mqtt/handler.go
```diff
@@ -370,6 +374,17 @@ func (h *handler) Disconnect(c *session.Client) {
 	}
 
 	h.logger.Info(fmt.Sprintf("client_id %s disconnected", c.ID))
+
+	if !c.WillFlag || c.CleanDisconnect {
+		return
+	}
+
+	if err := h.publishToBus(c, c.WillTopic, c.WillMessage); err != nil {
+		h.logger.Error(errors.Wrap(errFailedPublishWill, err).Error())
```
majabirmancevic: and here

duolok: There's a log on the lines under that with client id, no need there

duolok: Nvm, I guess we don't need a new error type

mqtt/handler.go
```diff
@@ -370,6 +373,17 @@ func (h *handler) Disconnect(c *session.Client) {
...
+	if !c.WillFlag || c.CleanDisconnect {
+		return
+	}
+
+	if err := h.publishToBus(c, c.WillTopic, c.WillMessage); err != nil {
+		h.logger.Error(fmt.Sprintf("client_id %s failed to publish will to topic %s: %s", c.ID, c.WillTopic, err))
+		return
+	}
+
+	h.logger.Info(fmt.Sprintf("client_id %s will published to topic %s", c.ID, c.WillTopic))
```
majabirmancevic: _.. published will message.._ maybe will be clearer?

mqtt/handler_test.go
```diff
@@ -420,6 +420,57 @@ func TestDisconnect(t *testing.T) {
...
+		}
+	}
+
+	cases := []struct {
+		desc          string
+		client        *session.Client
+		publishesWill bool
+	}{
+		{
+			desc:          "abnormal disconnect publishes the will",
```
majabirmancevic: can we replace `abnormal` with `unexpected` or something?

mqtt/handler.go
```diff
@@ -206,49 +206,52 @@ func (h *handler) Publish(c *session.Client, topic *string, payload *[]byte) {
 		return
 	}
 
+	if err := h.publishToBus(c, *topic, *payload); err != nil {
+		h.logger.Error(fmt.Sprintf("client_id %s failed to publish to topic %s: %s", c.ID, *topic, err))
```
majabirmancevic, line 210: there is `return errors.Wrap(messaging.ErrPublishMessage, err)` in `publishToBus` and then here we have, for example, "_client_id abc123 failed to publish to topic sensors/temp: failed to publish message: connection refused_";
let's try to avoid redundant messages

duolok: I don't see anything wrong, give an example of what you meant.

majabirmancevic: ^ this is an example: "client_id abc123 `failed to publish` to topic sensors/temp: `failed to publish` message: connection refused"
there is no need to duplicate this "wrapper message" in the log

mqtt/handler.go
```diff
@@ -206,49 +206,52 @@ func (h *handler) Publish(c *session.Client, topic *string, payload *[]byte) {
 		return
 	}
 
+	if err := h.publishToBus(c, *topic, *payload); err != nil {
+		h.logger.Error(fmt.Sprintf("client_id %s failed to publish to topic %s", c.ID, *topic))
```
majabirmancevic: missing `err`?

mqtt/handler.go
```diff
@@ -206,49 +206,52 @@ func (h *handler) Publish(c *session.Client, topic *string, payload *[]byte) {
...
 	tk := domain.ThingKey{
 		Value: string(c.Password),
 		Type:  c.Username,
 	}
 
 	pc, err := h.things.GetPubConfigByKey(context.Background(), tk)
 	if err != nil {
-		h.logger.Error(errors.Wrap(messaging.ErrPublishMessage, err).Error())
-		return
+		return errors.Wrap(messaging.ErrPublishMessage, err)
```
majabirmancevic: `return err`
no need to wrap it since we do it in callers, right?

mqtt/handler.go
majabirmancevic: related issue:https://github.com/MainfluxLabs/mainflux/issues/1281

mqtt/handler_test.go
```diff
@@ -298,14 +297,14 @@ func TestPublish(t *testing.T) {
 			client:  &sessionClient,
 			topic:   malformedSubtopics,
 			payload: payload,
-			logMsg:  messaging.ErrMalformedSubtopic.Error(),
+			logMsg:  "failed to publish to topic",
```
majabirmancevic: use `var`

majabirmancevic (approved):

manuio (approved):
LGTM

PR #1283, opened 2026-07-21
MF-1282 - drop dbutil.GetOrderQuery and handle LOWER per-service

### LINKED ISSUE #1282: security: sanitize ORDER BY and IN clause inputs in dbutil to prevent SQL injection
### Description
`dbutil.GetOrderQuery()` in `pkg/dbutil/query.go` returned unvalidated user input directly into raw `fmt.Sprintf("... ORDER BY %s ...")` SQL strings via its `default` switch case. Any unsanitized string passed via API pagination query parameters (`?order=...`) could lead to SQL injection.

Additionally, `GetGroupIDsQuery` in `pkg/dbutil/query.go` and `getIDsQuery` in `things/postgres/profiles.go` formatted string slices into SQL `IN ('...')` clauses using `strings.Join` without single-quote escaping.

### Expected Behavior
- `GetOrderQuery` should strictly validate column identifiers matching `^[a-zA-Z0-9_.]+$`, falling back to `"id"` for any input containing unsafe characters.
- `GetGroupIDsQuery` and `getIDsQuery` should sanitize single quotes in input strings.

### Steps to Reproduce
Call `dbutil.GetOrderQuery("id; DROP TABLE users; --")`. Previously returned `"id; DROP TABLE users; --"`.

pranav718:
### What does this do?
- **Dropped `dbutil.GetOrderQuery`**: Removed redundant `GetOrderQuery` helper from `pkg/dbutil/query.go`. Handled `LOWER()` case-insensitive ordering (`name`, `email`) per-service inside repository packages where needed.
- **Reverted `IN` Clause Escaping**: Reverted single-quote escaping in `GetGroupIDsQuery` and `getIDsQuery` to preserve clean string joining without defensive DB-layer escaping.
- **Unit Tests**: Added `pkg/dbutil/query_test.go` with unit tests covering `GetGroupIDsQuery`, `GetDirQuery`, `GetNameQuery`, `GetOffsetLimitQuery`, and `BuildWhereClause`.

Resolves #1282

### List any changes that modify/break current functionality
None. All valid sorting options continue to work normally.

Yes, added unit tests in `pkg/dbutil/query_test.go`.

N/A

### Notes
None.

### DIFF
--- a/audit/postgres/events.go
+++ b/audit/postgres/events.go
@@ -65,7 +65,7 @@ func (r *eventRepository) RetrieveEvents(ctx context.Context, pm audit.PageMetad
 	whereClause := dbutil.BuildWhereClause(actorEmailQ, opQ, fromQ, toQ, dataQ)
 	query := fmt.Sprintf(
 		`SELECT id, occurred_at, operation, actor_id, actor_email, org_id, group_id, action_data FROM events %s ORDER BY %s %s %s`,
-		whereClause, dbutil.GetOrderQuery(pm.Order), dbutil.GetDirQuery(pm.Dir), dbutil.GetOffsetLimitQuery(pm.Limit),
+		whereClause, dbutil.GetOrderQuery(pm.Order, audit.EventOrderFields), dbutil.GetDirQuery(pm.Dir), dbutil.GetOffsetLimitQuery(pm.Limit),
 	)
 	cquery := fmt.Sprintf(`SELECT COUNT(*) FROM events %s`, whereClause)
 
@@ -97,7 +97,7 @@ func (r *eventRepository) RetrieveEventsByOrg(ctx context.Context, orgID string,
 	whereClause := dbutil.BuildWhereClause(actorEmailQ, opQ, orgQ, fromQ, toQ, dataQ)
 	query := fmt.Sprintf(
 		`SELECT id, occurred_at, operation, actor_id, actor_email, org_id, group_id, action_data FROM events %s ORDER BY %s %s %s`,
-		whereClause, dbutil.GetOrderQuery(pm.Order), dbutil.GetDirQuery(pm.Dir), dbutil.GetOffsetLimitQuery(pm.Limit),
+		whereClause, dbutil.GetOrderQuery(pm.Order, audit.EventOrderFields), dbutil.GetDirQuery(pm.Dir), dbutil.GetOffsetLimitQuery(pm.Limit),
 	)
 	cquery := fmt.Sprintf(`SELECT COUNT(*) FROM events %s`, whereClause)
 
@@ -130,7 +130,7 @@ func (r *eventRepository) RetrieveEventsByGroup(ctx context.Context, groupID str
 	whereClause := dbutil.BuildWhereClause(actorEmailQ, opQ, groupQ, fromQ, toQ, dataQ)
 	query := fmt.Sprintf(
 		`SELECT id, occurred_at, operation, actor_id, actor_email, org_id, group_id, action_data FROM events %s ORDER BY %s %s %s`,
-		whereClause, dbutil.GetOrderQuery(pm.Order), dbutil.GetDirQuery(pm.Dir), dbutil.GetOffsetLimitQuery(pm.Limit),
+		whereClause, dbutil.GetOrderQuery(pm.Order, audit.EventOrderFields), dbutil.GetDirQuery(pm.Dir), dbutil.GetOffsetLimitQuery(pm.Limit),
 	)
 	cquery := fmt.Sprintf(`SELECT COUNT(*) FROM events %s`, whereClause)
 
--- a/auth/postgres/invites.go
+++ b/auth/postgres/invites.go
@@ -307,7 +307,7 @@ func (ir invitesRepository) RetrieveOrgInvitesByOrg(ctx context.Context, orgID s
 	filterNonDormant := `invitee_id IS NOT NULL`
 
 	whereClause := dbutil.BuildWhereClause(filterOrgID, filterState, filterNonDormant)
-	oq := dbutil.GetOrderQuery(pm.Order)
+	oq := dbutil.GetOrderQuery(pm.Order, auth.OrgInviteOrderFields)
 	dq := dbutil.GetDirQuery(pm.Dir)
 	olq := dbutil.GetOffsetLimitQuery(pm.Limit)
 
@@ -384,7 +384,7 @@ func (ir invitesRepository) RetrieveOrgInvitesByUser(ctx context.Context, userTy
 	filterNonDormant := `invitee_id IS NOT NULL`
 
 	whereClause := dbutil.BuildWhereClause(filterUserType, filterState, filterNonDormant)
-	oq := dbutil.GetOrderQuery(pm.Order)
+	oq := dbutil.GetOrderQuery(pm.Order, auth.OrgInviteOrderFields)
 	dq := dbutil.GetDirQuery(pm.Dir)
 	olq := dbutil.GetOffsetLimitQuery(pm.Limit)
 
--- a/consumers/alarms/postgres/alarms.go
+++ b/consumers/alarms/postgres/alarms.go
@@ -106,7 +106,7 @@ func (ar *alarmRepository) RetrieveByThing(ctx context.Context, thingID string, 
 		return alarms.AlarmsPage{}, errors.Wrap(dbutil.ErrNotFound, err)
 	}
 
-	oq := dbutil.GetOrderQuery(pm.Order)
+	oq := dbutil.GetOrderQuery(pm.Order, alarms.AlarmOrderFields)
 	dq := dbutil.GetDirQuery(pm.Dir)
 	olq := dbutil.GetOffsetLimitQuery(pm.Limit)
 
@@ -137,7 +137,7 @@ func (ar *alarmRepository) RetrieveByGroup(ctx context.Context, groupID string, 
 		return alarms.AlarmsPage{}, errors.Wrap(dbutil.ErrNotFound, err)
 	}
 
-	oq := dbutil.GetOrderQuery(pm.Order)
+	oq := dbutil.GetOrderQuery(pm.Order, alarms.AlarmOrderFields)
 	dq := dbutil.GetDirQuery(pm.Dir)
 	olq := dbutil.GetOffsetLimitQuery(pm.Limit)
 
@@ -168,7 +168,7 @@ func (ar *alarmRepository) RetrieveByGroups(ctx context.Context, groupIDs []stri
 		return alarms.AlarmsPage{}, nil
 	}
 
-	oq := dbutil.GetOrderQuery(pm.Order)
+	oq := dbutil.GetOrderQuery(pm.Order, alarms.AlarmOrderFields)
 	dq := dbutil.GetDirQuery(pm.Dir)
 	olq := dbutil.GetOffsetLimitQuery(pm.Limit)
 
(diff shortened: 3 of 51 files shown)


consumers/alarms/postgres/alarms.go
```diff
@@ -376,3 +376,12 @@ func timeRangeQuery(from, to int64) string {
 	}
 	return strings.Join(parts, " AND ")
 }
+
+func getLowerOrder(order string) string {
+	switch order {
+	case "name", "email":
```
majabirmancevic: There is no `name` column in the `alarms` table. Please check all services where we should use `LOWER` for their fields (we don't need `email` case everywhere), and let's define only the fields that a certain service needs.

pranav718: Thanks @majabirmancevic  , Just audited all the service repos and updated them:

- Dropped `getLowerOrder` completely from services without text order fields (`alarms`, `keys`, `invites`).
- Tailored `getLowerOrder` in each repo to only handle their specific text columns (`name`, `url`, `email`, etc.).

Pushing the update now

audit/postgres/events.go
```diff
@@ -295,3 +295,12 @@ func nullableString(s string) sql.NullString {
 	}
 	return sql.NullString{String: s, Valid: true}
 }
+
+func getLowerOrder(order string) string {
+	switch order {
+	case "name", "email":
```
manuio: Audit service is using `actor_email` not `email`. BTW, it looks like we are wrongly using `email` in certain places.

manuio (changes requested):

audit/postgres/events.go
pranav718: Thanks @manuio, updating `events.go` to check `actor_email` specifically.

also yes, I previously went through all the other service repos and cleaned up those extra `email` cases,  `email` is now only kept in `users`, and `getLowerOrder` has been dropped completely from tables with no text sorting (like `alarms` & `invites`).

manuio: I meant that in audit service we are using `email` as parameter in several places like here: https://github.com/MainfluxLabs/mainflux/blob/master/audit/postgres/events.go#L73

auth/postgres/invites.go
```diff
@@ -307,7 +307,7 @@ func (ir invitesRepository) RetrieveOrgInvitesByOrg(ctx context.Context, orgID s
 	filterNonDormant := `invitee_id IS NOT NULL`
 
 	whereClause := dbutil.BuildWhereClause(filterOrgID, filterState, filterNonDormant)
-	oq := dbutil.GetOrderQuery(pm.Order)
+	oq := pm.Order
```
manuio: Use `getLowerOrder`

manuio (changes requested):

audit/postgres/events.go
majabirmancevic: > I meant that in audit service we are using email as parameter in several places like here: https://github.com/MainfluxLabs/mainflux/blob/master/audit/postgres/events.go#L73

Here is the related issue https://github.com/MainfluxLabs/mainflux/issues/1287

users/postgres/users.go
```diff
@@ -334,3 +334,12 @@ func createEmailQuery(entity string, email string) (string, string, error) {
 
 	return query, param, nil
 }
+
+func getLowerOrder(order string) string {
+	switch order {
+	case "name", "email":
```
majabirmancevic: do we need "name" here?

consumers/notifiers/postgres/notifiers.go
```diff
@@ -246,3 +246,12 @@ func toNotifier(dbN dbNotifier) (notifiers.Notifier, error) {
 		Metadata: metadata,
 	}, nil
 }
+
+func getLowerOrder(order string) string {
+	switch order {
+	case "name", "contacts":
```
majabirmancevic: `contacts` is a real column, but `allowedOrders` only exposes `id`, `name`; so this case is unreachable via any validated HTTP request

downlinks/postgres/downlinks.go
```diff
@@ -405,3 +405,12 @@ func toDownlink(dbD dbDownlink) (downlinks.Downlink, error) {
 		Metadata:   metadata,
 	}, nil
 }
+
+func getLowerOrder(order string) string {
+	switch order {
+	case "name", "url":
```
majabirmancevic: same here for `url`, `allowedOrders` only has `id`, `name`

webhooks/postgres/webhooks.go
```diff
@@ -312,3 +312,12 @@ func toWebhook(dbW dbWebhook) (webhooks.Webhook, error) {
 		Metadata: metadata,
 	}, nil
 }
+
+func getLowerOrder(order string) string {
+	switch order {
+	case "name", "url":
```
majabirmancevic: `allowedOrders` only has `id`, `name`

audit/postgres/events.go
pranav718: Ah got it, thanks for clarifying! I'll leave that for #1287 then.

auth/postgres/invites.go
pranav718: makes sense, will update!

users/postgres/users.go
pranav718: will check allowedOrders and update accordingly.

pranav718:
Updated based on all the review feedback:

- Tailored `getLowerOrder` per service to only match fields from `allowedOrders` (removed `url`, `contacts`, `name` where not applicable)
- Changed `users` to only handle `email` (no `name` in `allowedOrders`)
- Using `getLowerOrder(pm.Order)` everywhere instead of raw `pm.Order`

Thanks for the thorough reviews @majabirmancevic @manuio!

consumers/alarms/postgres/alarms.go
```diff
@@ -376,3 +376,7 @@ func timeRangeQuery(from, to int64) string {
 	}
 	return strings.Join(parts, " AND ")
 }
+
+func getLowerOrder(order string) string {
```
majabirmancevic: there is no need for this helper

pranav718: got it, will be using `pm.Order` directly

majabirmancevic:
Hey, we looked at this more and want to propose a different approach before merging. 

 We'd like one shared, generic resolver that defaults to a safe column instead of passing through, plus per-table maps instead of per-service ones (a few of the existing `allowedOrders` maps are already shared across tables with different schemas, e.g. auth's spans orgs/org_invites/keys, which don't have the same columns).

One new function, does all the work:

```
// pkg/dbutil/order.go
package dbutil

import "fmt"

// GetOrderQuery resolves order against fields, which maps API-facing order
// keys to real SQL column expressions (with LOWER() baked in where needed).
// Falls back to "id" for anything not in fields, so this is safe even if a
// caller forgets to validate order upstream.
func GetOrderQuery(order string, fields map[string]string) string {
	if col, ok := fields[order]; ok {
		return col
	}
	return "id"
}
```
This replaces dbutil.GetOrderQuery (old) and every per-file getLowerOrder.

Per-table maps, colocated with the domain files (same place the sentinel errors already live, e.g. ErrOrgNotEmpty in auth/orgs.go) so both `api/ `and `postgres/` can import them without adding a new dependency:


```
// auth/orgs.go
var OrgOrderFields = map[string]string{
	"id":         "id",
	"name":       "LOWER(name)",
	"created_at": "created_at",
	"updated_at": "updated_at",
}

// auth/invites.go
var OrgInviteOrderFields = map[string]string{
	"id":         "id",
	"invitee_id": "invitee_id",
	"inviter_id": "inviter_id",
	"org_id":     "org_id",
	"state":      "state",
	"created_at": "created_at",
}

// auth/key.go
var KeyOrderFields = map[string]string{
	"id":        "id",
	"issued_at": "issued_at",
}

// audit/events.go 
var EventOrderFields = map[string]string{
	"id":          "id",
	"occurred_at": "occurred_at",
	"operation":   "operation",
	"actor_email": "LOWER(actor_email)",
	"org_id":      "org_id",
	"group_id":    "group_id",
}

// things/things.go
var ThingOrderFields = map[string]string{
	"id":   "id",
	"name": "LOWER(name)",
	"type": "type",
}

// things/groups.go
var GroupOrderFields = map[string]string{
	"id":         "id",
	"name":       "LOWER(name)",
	"created_at": "created_at",
	"updated_at": "updated_at",
}

// things/profiles.go
var ProfileOrderFields = map[string]string{
	"id":   "id",
	"name": "LOWER(name)",
}

// consumers/notifiers/notifiers.go
var NotifierOrderFields = map[string]string{
	"id":       "id",
	"name":     "LOWER(name)",
}

// downlinks/downlinks.go
var DownlinkOrderFields = map[string]string{
	"id":   "id",
	"name": "LOWER(name)",
}

// webhooks/webhooks.go
var WebhookOrderFields = map[string]string{
	"id":   "id",
	"name": "LOWER(name)",
}

// modbus/clients.go
var ClientOrderFields = map[string]string{
	"id":   "id",
	"name": "LOWER(name)",
}

// rules/rules.go 
var RuleOrderFields = map[string]string{
	"id":   "id",
	"name": "LOWER(name)",
}

// consumers/alarms/alarms.go
var AlarmOrderFields = map[string]string{
	"id":      "id",
	"created": "created",
	"level":   "level",
	"status":  "status",
}
One exception: users.UserOrderFields should go in pkg/domain/users.go instead of the users root package, because auth's org-memberships endpoint needs to validate/resolve order for a query that actually hits the users table via gRPC, and auth shouldn't import users directly — it already imports pkg/domain for domain.UsersPageMetadata.

// pkg/domain/users.go
var UserOrderFields = map[string]string{
	"id":    "id",
	"email": "LOWER(email)",
}

// users/invites.go — platform_invites table
var PlatformInviteOrderFields = map[string]string{
	"id":            "id",
	"invitee_email": "LOWER(invitee_email)",
	"state":         "state",
	"created_at":    "created_at",
}
```

Then each `api/validation.go` checks `_, ok := X.OrderFields[pm.Order]`, and each `postgres/*.go` calls `dbutil.GetOrderQuery(pm.Order, X.OrderFields)` - one source of truth per table, no drift possible between what's "valid" and what SQL gets built.

Let us know if you want to take this, or if you'd rather we take it from here.

majabirmancevic:
Also, merging this PR is blocked due to failing tests.

pranav718:
Thanks for the detailed proposal, this is a really clean approach and I'd like to take this on.

So just to confirm:
- One new `GetOrderQuery(order, fields)` in `pkg/dbutil` with a safe fallback to `"id"`
- Per-table `OrderFields` maps in the domain files
- `UserOrderFields` in `pkg/domain/users.go` to avoid circular import
- Both api validation and postgres layers use the same maps as single source of truth

Will rework the PR with the new approach!

majabirmancevic:
That's right! :)
Thank you, and sorry for this extra work.

pranav718:
No worries at all, on it!

manuio (dismissed):
LGTM

manuio:
@pranav718 Solid fix overall — could you add test cases for GetGroupIDsQuery, GetDirQuery, GetNameQuery, GetOffsetLimitQuery, and BuildWhereClause too, since the description mentions them but the test file currently only covers GetOrderQuery?

pranav718:
Added tests for GetDirQuery, GetNameQuery, GetOffsetLimitQuery, GetGroupIDsQuery, and BuildWhereClause, all passing!

manuio (approved):
LGTM

manuio:
@pranav718 Thank you for the addition! Just a remark, it's better to add new commits to the history and then let the maintainer squash commits, than force pushing. Otherwise, you can't see the diff per commit.

pranav718:
I'll keep that in mind for future contributions. Thanks for the feedback and for merging!

PR #1285, opened 2026-07-22
MF-1241 - Use protomfx.Command in adapers

### LINKED ISSUE #1241: Consider using PublishCommand in adapters and use mfxproto.Commands throughout the codebase
Currently, we are only using Commands struct in certs scheduler. We should use Commands in http, coap and ws adapters. 

```

func sendCommandToThingEndpoint(svc http.Service) endpoint.Endpoint {
	return func(ctx context.Context, request any) (any, error) {
		req := request.(thingCommandReq)

		if err := req.validate(); err != nil {
			return nil, err
		}

		if req.thingKey.Value != "" {
			return nil, svc.SendCommandToThingByKey(ctx, req.thingKey, req.id, req.msg)
		}

		return nil, svc.SendCommandToThing(ctx, req.token, req.id, req.msg)
	}
}
```
```

func (as *adapterService) SendCommandToThing(ctx context.Context, token, thingID string, message protomfx.Message) error {
	if err := as.things.CanUserAccessThing(ctx, domain.UserAccessReq{Token: token, ID: thingID, Action: domain.GroupEditor}); err != nil {
		return err
	}

	if err := as.publisher.Publish(nats.GetThingCommandsSubject(thingID, message.Subtopic), message); err != nil {
		return err
	}

	return nil
}
```

These methods should be refactored.

majabirmancevic:
Resolves #1241

### DIFF
--- a/coap/api/transport.go
+++ b/coap/api/transport.go
@@ -88,14 +88,14 @@ func handler(w mux.ResponseWriter, m *mux.Message) {
 	case codes.GET:
 		err = handleGet(m, w.Client(), key)
 	case codes.POST:
-		msg, decErr := decodeMessage(m)
+		payload, decErr := readPayload(m)
 		if decErr != nil {
 			logger.Warn(fmt.Sprintf("Error decoding message: %s", decErr))
 			resp.Code = codes.BadRequest
 			sendResp(w, &resp)
 			return
 		}
-		err = handlePost(m, msg, key)
+		err = handlePost(m, payload, key)
 	default:
 		err = dbutil.ErrNotFound
 	}
@@ -115,7 +115,7 @@ func handler(w mux.ResponseWriter, m *mux.Message) {
 	}
 }
 
-func handlePost(m *mux.Message, msg protomfx.Message, key domain.ThingKey) error {
+func handlePost(m *mux.Message, payload []byte, key domain.ThingKey) error {
 	path, err := m.Options.Path()
 	if err != nil {
 		return errBadOptions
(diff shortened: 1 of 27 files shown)


ws/mocks/pubsub.go
```diff
@@ -8,16 +8,23 @@ import (
...
 )
 
-var _ messaging.PubSub = (*mockPubSub)(nil)
+var _ nats.PubSub = (*mockPubSub)(nil)
 
 type MockPubSub interface {
 	Publish(string, protomfx.Message) error
 	Subscribe(string, string, messaging.MessageHandler) error
 	Unsubscribe(string, string) error
+	PublishAlarm(string, protomfx.Alarm) error
```
majabirmancevic, line 22: `ws.PubSub` only needs `Publish`/`Subscribe`/`Unsubscribe`/`PublishCommand`. Left as-is here to keep this PR scoped to the Command work; find a way to trim it in a follow-up (for other adapters as well).

pkg/messaging/mapper.go
```diff
@@ -0,0 +1,117 @@
...
+		Payload:     payload,
+	}
+}
+
+// extractCreated unwraps a "Created" field embedded in the payload itself,
+// preferring it over the proto-level Created timestamp.
+func extractCreated(payload []byte, created int64) (int64, []byte) {
+	var p map[string]any
+
+	if len(payload) > 0 {
```
manuio: Can we check`if len(payload) < 0 {` and `return created, payload`?

coap/api/transport.go
```diff
@@ -88,14 +88,14 @@ func handler(w mux.ResponseWriter, m *mux.Message) {
 	case codes.GET:
 		err = handleGet(m, w.Client(), key)
 	case codes.POST:
-		msg, decErr := decodeMessage(m)
+		payload, decErr := readPayload(m)
```
manuio, line 91: Do we need `readPayload`? How we can get an error here?

manuio (changes requested):

coap/api/transport.go
majabirmancevic:
```
func readPayload(msg *mux.Message) ([]byte, error) {
	if msg.Body == nil {
		return []byte{}, nil
	}
	return io.ReadAll(msg.Body)
}
```

yes, `Body` is nil when there's no payload, and `io.ReadAll(nil)` would panic, so we need the nil check;
in practice, the error return itself is unlikely to ever fire, but keeping it just honors `io.Reader`'s contract (`func io.ReadAll(r io.Reader) ([]byte, error)`)

manuio: ok, but the helper is only used in a single place right?

majabirmancevic: yes, as before with `decodeMessage`

ws/api/endpoints.go
```diff
@@ -133,3 +133,12 @@ func buildMessage(subtopic string, payload []byte) protomfx.Message {
...
 }
+
+func buildCommand(subtopic string, payload []byte) protomfx.Command {
+	return protomfx.Command{
+		Subtopic: subtopic,
+		Protocol: protocol,
+		Payload:  payload,
+		Created:  time.Now().UnixNano(),
+	}
+}
```
duolok, line 144: Either use `buildCommand` or `newCommand`

coap/api/transport.go
```diff
@@ -171,22 +173,29 @@ func handleGet(m *mux.Message, c mux.Client, key domain.ThingKey) error {
...
-		}
-		ret.Payload = buff
+func newCommand(subtopic string, payload []byte) protomfx.Command {
+	return protomfx.Command{
+		Protocol: protocol,
+		Subtopic: subtopic,
+		Payload:  payload,
+		Created:  time.Now().UnixNano(),
 	}
+}
```
duolok: And here

duolok (changes requested):

ws/api/endpoints.go
majabirmancevic: fixed in ws to use `buildCommand`

duolok (approved):
LGTM

manuio (approved):
LGTM

PR #1291, opened 2026-07-27
MF-1290 - use transaction tx instead of db in groupMembershipsReposit…

### LINKED ISSUE #1290: Bypass of SQL transaction in things group memberships Save
In `things/postgres/memberships.go`, the `Save` method starts a database transaction via `mr.db.BeginTxx(ctx, nil)` and sets up `defer tx.Rollback()`, but the subsequent `INSERT` statements inside the loop execute against `mr.db` (the direct connection pool) instead of `tx`:
https://github.com/MainfluxLabs/mainflux/blob/master/things/postgres/memberships.go#L28-L53

```go
func (mr groupMembershipsRepository) Save(ctx context.Context, gms ...things.GroupMembership) error {
	tx, err := mr.db.BeginTxx(ctx, nil)
	if err != nil {
		return err
	}
	defer tx.Rollback()
	q := `INSERT INTO group_memberships (member_id, group_id, role) VALUES (:member_id, :group_id, :role);`
	for _, g := range gms {
		dbgm := toDBGroupMembership(g)
		if _, err := mr.db.NamedExecContext(ctx, q, dbgm); err != nil { // bug: uses mr.db instead of tx
			// ...
		}
	}
	if err = tx.Commit(); err != nil {
		return errors.Wrap(dbutil.ErrCreateEntity, err)
	}
	return nil
}
```

**Impact**
Because mr.db is used inside the loop instead of tx:

The transaction (tx) remains empty and its tx.Commit() / tx.Rollback() calls do nothing.
If a batch insert of multiple group memberships fails mid-loop (e.g. on item 3 of 5), the previously processed items (1 and 2) are committed permanently directly to the database, violating atomicity and leaving the database in a partially-saved state.
Sibling repositories (things/postgres/profiles.go, things/postgres/groups.go, auth/postgres/memberships.go) all correctly execute queries against tx.

**Proposed Fix**
Change mr.db.NamedExecContext to tx.NamedExecContext on line 39 of things/postgres/memberships.go.

pranav718:
### What does this do?
Fixes transaction bypass in `groupMembershipsRepository.Save` where `mr.db.NamedExecContext` was mistakenly used inside the batch loop instead of `tx.NamedExecContext`.

Resolves #1290

### List any changes that modify/break current functionality
None.

N/A

N/A

### Notes
None.

### DIFF
--- a/things/postgres/memberships.go
+++ b/things/postgres/memberships.go
@@ -36,7 +36,7 @@ func (mr groupMembershipsRepository) Save(ctx context.Context, gms ...things.Gro
 
 	for _, g := range gms {
 		dbgm := toDBGroupMembership(g)
-		if _, err := mr.db.NamedExecContext(ctx, q, dbgm); err != nil {
+		if _, err := tx.NamedExecContext(ctx, q, dbgm); err != nil {
 			pgErr, ok := err.(*pgconn.PgError)
 			if ok {
 				switch pgErr.Code {


majabirmancevic (approved):

manuio (approved):
LGTM

PR #1294, opened 2026-07-28
MF-1293 - Dedupe JSON messages by payload hash

### LINKED ISSUE #1293: Prevent duplicate JSON messages from being stored on repeated downlink fetches
The postgres/timescale json writer has no way to distinguish an exact re-delivered message (e.g. a downlink fetch that returns the same payload twice) from a new one, so redeliveries were stored as duplicate rows.

majabirmancevic:
Added a `payload_hash` column (FNV-1a 32-bit hash of the payload) and a unique index on `(created, publisher, subtopic, payload_hash)`, with `INSERT ... ON CONFLICT DO NOTHING`. 
An exact duplicate payload at the same key is silently dropped by the DB; different payload content at the same key is still stored. 
Applied to both the postgres and timescale `json` writers/readers; `senml` is unaffected.

Resolves #1293

### DIFF
--- a/consumers/writers/postgres/consumer.go
+++ b/consumers/writers/postgres/consumer.go
@@ -5,7 +5,6 @@ package postgres
 
 import (
 	"context"
-	"hash/fnv"
 
 	"github.com/MainfluxLabs/mainflux/consumers"
 	"github.com/MainfluxLabs/mainflux/pkg/errors"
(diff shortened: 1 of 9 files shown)


consumers/writers/postgres/consumer.go
```diff
@@ -92,9 +93,15 @@ func (pr postgresRepo) saveSenML(msgs []protomfx.Message) (err error) {
...
 
+type jsonRow struct {
+	mfjson.Message
+	PayloadHash int32 `db:"payload_hash"`
+}
+
 func (pr postgresRepo) saveJSON(msgs []protomfx.Message) error {
-	q := `INSERT INTO json (created, subtopic, publisher, protocol, payload)
-          VALUES (:created, :subtopic, :publisher, :protocol, :payload);`
+	q := `INSERT INTO json (created, subtopic, publisher, protocol, payload, payload_hash)
```
manuio, line 102: Why are we using `payload_hash` if we do nothing on conflict?

manuio (changes requested):

consumers/writers/postgres/consumer.go
majabirmancevic: `payload_hash` in the index answers "what counts as a duplicate", while `ON CONFLICT () DO NOTHING` answers "when we hit one, skip that row instead of erroring out the whole batch"

anna02272 (approved):

manuio (approved):
LGTM

Issue #1204, opened 2026-04-21
Clean up consumer/message handler types and interfaces


bool3max:
`pkg/messaging.MessageHandler` and `consumers.MessageConsumer` (as well as the analogous alarm-related types) are very similar. Consumer code mixes the `messaging.Subscriber` (which accept the aforementioned `MessageHandler` type) and `AlarmConsumer`/`MessageConsumer` interfaces, which requires the existence of the seemingly useless `consumers.messageAdapter` and `consumers.alarmAdapter` types.

Also, `consumers.Subscription` is seemingly unnecessary and introduces unneeded complexity. `consumers.Message()` and `consumers.Alarms()` should be able to start the subscribers themselves without the need for an additional `consumers.Start(...)`.

See if these can be cleaned up / simplified.

PR #1296, opened 2026-07-30
MF-1244 - Introduce protomfx.Notification and protomfx.Webhook 

### LINKED ISSUE #1244: Introduce proto structs for notifiers and webhooks

majabirmancevic:
New `Notification` and `Webhook` proto structures for publish/subscribe on NATS have been introduced.

Resolves #1244


pkg/messaging/nats/pubsub.go
```diff
@@ -44,13 +45,20 @@ const (
...
 
-// PubSub extends messaging.PubSub with alarm/command publishing, and alarm/command subscribing.
+// PubSub extends messaging.PubSub with alarm/command/notification/webhook publishing and subscribing.
 type PubSub interface {
 	messaging.PubSub
 	messaging.AlarmPublisher
 	messaging.AlarmSubscriber
 	messaging.CommandPublisher
 	messaging.CommandSubscriber
+	messaging.NotificationPublisher
```
majabirmancevic, line 55: This PR follows the existing `Alarm`/`Command` pattern exactly for `Notification` and `Webhook` (explicit per-type `Publisher`/`Handler`/`Subscriber` in `pkg/messaging` + matching nats package implementations). That keeps this PR consistent and low-risk, but it does mean `pkg/messaging/pubsub.go` and `pkg/messaging/nats/{publisher,pubsub}.go` now carry four near-identical sets of interfaces and handler boilerplate instead of two.

Deferring that cleanup intentionally - it overlaps with #1204, which already tracks reducing duplication in this messaging layer. A follow-up PR should revisit the pub/sub mechanism so adding a future proto struct (a 5th typed stream) doesn't require repeating this same boilerplate again.

manuio (approved):
LGTM

anna02272 (approved):

PR #1299, opened 2026-08-11
MF-958 - Reduce readers struct duplication

### LINKED ISSUE #958: Polish readers API metadata struct definitions
The `SenMLPageMetadata` and `JSONPageMetadata` structs share a subset of fields:

```go
// SenMLPageMetadata represents the parameters used to create database queries
type SenMLPageMetadata struct {
	Offset      uint64  `json:"offset"`
	Limit       uint64  `json:"limit"`
	Subtopic    string  `json:"subtopic,omitempty"`
	Publisher   string  `json:"publisher,omitempty"`
	Protocol    string  `json:"protocol,omitempty"`
	Name        string  `json:"name,omitempty"`
	Value       float64 `json:"v,omitempty"`
	Comparator  string  `json:"comparator,omitempty"`
	BoolValue   bool    `json:"vb,omitempty"`
	StringValue string  `json:"vs,omitempty"`
	DataValue   string  `json:"vd,omitempty"`
	From        int64   `json:"from,omitempty"`
	To          int64   `json:"to,omitempty"`
	AggInterval string  `json:"agg_interval,omitempty"`
	AggValue    uint64  `json:"agg_value,omitempty"`
	AggType     string  `json:"agg_type,omitempty"`
	AggField    string  `json:"agg_field,omitempty"`
	Dir         string  `json:"dir,omitempty"`
}

// JSONPageMetadata represents the parameters used to create database queries
type JSONPageMetadata struct {
	Offset      uint64 `json:"offset"`
	Limit       uint64 `json:"limit"`
	Subtopic    string `json:"subtopic,omitempty"`
	Publisher   string `json:"publisher,omitempty"`
	Protocol    string `json:"protocol,omitempty"`
	From        int64  `json:"from,omitempty"`
	To          int64  `json:"to,omitempty"`
	Filter      string `json:"filter,omitempty"`
	AggInterval string `json:"agg_interval,omitempty"`
	AggValue    uint64 `json:"agg_value,omitempty"`
	AggType     string `json:"agg_type,omitempty"`
	AggField    string `json:"agg_field,omitempty"`
	Dir         string `json:"dir,omitempty"`
}
```

See if they can be moved to a common struct or find another way to minimize duplication.

duolok:
Resolves #958


pkg/domain/readers.go
```diff
@@ -26,19 +26,13 @@ type SenMLMessagesPage struct {
 	MessagesPage
 }
 
-// SenMLPageMetadata represents the parameters used to create database queries.
-type SenMLPageMetadata struct {
+// ReadersMetadata represents the query parameters shared by all message formats.
```
majabirmancevic: should we use `PageMetadata` instead of just `Metadata`?

duolok: I was considering that, but we never return `ReadersPageMetadata`, so I guess shorter name works as well.

readers/timescale/common.go
```diff
@@ -0,0 +1,52 @@
...
+
+const (
+	jsonTable  = "json"
+	jsonOrder  = "created"
+	senmlTable = "senml"
+	senmlOrder = "time"
+)
+
+// baseConditions returns the SQL predicates shared by all message formats.
+func baseConditions(pm readers.ReadersMetadata, timeColumn string) []string {
```
majabirmancevic: Are `postgres/common.go` and `timescale/common.go` the same? Can we move them to `pkg` and use them from there?

duolok: I agree.

pkg/domain/readers.go
majabirmancevic: hmm, or `PageMeta`?

manuio: In any case I wouldn't rename in the same PR

majabirmancevic: ok

duolok: I think `PageMeta` is too general, we could have that name in any service and we wouldn't know that the struct contains only fields for readers. For example, you can use `PageMeta` in `things`, `rules` etc. Maybe some other name? Also what's wrong with `ReadersMetadata`?

majabirmancevic: I meant using `Page` as a prefix for `Metadata`, so `ReadersMetadata` -> `ReadersPageMetadata` etc..

duolok: I see. I mean we could use `ReadersPageMetadata` but my reasoning is that we use `PageMetadata` only for types that we at some point return. That's why I didn't use it and also it seems like a long name to type. But maybe we could rename

majabirmancevic: ok I get it, but maybe we should make a distinction between `PageMetadata` and `Metadata` since we use both terms in the codebase

manuio: `ReadersParams`?

pkg/readers/query.go
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) Mainflux
+// SPDX-License-Identifier: Apache-2.0
+
+package readers
```
majabirmancevic, line 4: `common.go` could be renamed to `query.go`

readers/api/grpc/server.go
```diff
@@ -60,24 +60,45 @@ func (gs *grpcServer) ListSenMLMessages(ctx context.Context, req *protomfx.ListS
...
+	GetFrom() int64
+	GetTo() int64
+	GetAggInterval() string
+	GetAggValue() uint64
+	GetAggType() string
+	GetAggFields() []string
+	GetDir() string
+}
+
+func decodeReadersMetadata(req readersMetadataReq) domain.ReadersParams {
```
majabirmancevic: fix naming

readers/timescale/senml.go
```diff
@@ -187,64 +187,43 @@ func (sr *senmlRepository) scanMessages(rows *sqlx.Rows) ([]readers.Message, err
...
 }
 
-func (sr *senmlRepository) fmtCondition(rpm readers.SenMLPageMetadata) string {
-	var query map[string]any
-	meta, err := json.Marshal(rpm)
-	if err != nil {
-		return ""
+// senmlConditions returns the SQL predicates common to every senml read,
+// including the aggregated ones.
+func senmlConditions(rpm readers.SenMLPageMetadata) []string {
```
majabirmancevic: does it make sense to move this helper to `pkg` as well, since we have the same helper in `/postgres` and `/timescale`?

majabirmancevic (dismissed):

manuio (approved):
LGTM

majabirmancevic (approved):

PR #1302, opened 2026-08-12
MF-1298 - Remove support for RabbitMQ message broker

### LINKED ISSUE #1298: Consider removing support for RabbitMQ
We haven't used RabbitMQ in practice for some time. Only `postgres-writer` and `timescale-writer` go through the broker-agnostic layer that supports it - every other service already hardcodes NATS directly. Since switching the internal message bus to NATS JetStream (#1272), NATS now handles the durable, guaranteed-delivery use case that RabbitMQ originally provided, so there's no remaining functional reason to keep a second broker implementation around. 
`pkg/messaging/rabbitmq` hasn't been updated since before that switch and is not wired into either deployment path we ship (Helm chart or docker-compose).

majabirmancevic:
Resolves #1298


duolok (approved):
LGTM

manuio (approved):
LGTM

PR #1304, opened 2026-08-14
MF-1303 - Replace nats.Publisher/nats.PubSub with point-of-use interfaces

### LINKED ISSUE #1303: Replace nats.Publisher/nats.PubSub with point-of-use interfaces
Services that talk to the message broker (`http`, `mqtt`, `ws`, `coap`, `rules`, `converters`, `modbus`, `downlinks`) import `pkg/messaging/nats` directly and depend on its big `Publisher`/`PubSub` interfaces, even though each only uses a handful of their methods. Have each service declare its own minimal interface for what it actually needs.

majabirmancevic:
Each service now declares the minimal publish interface it actually needs, rather than depending on the full `nats.Publisher/nats.PubSub`. This decouples business logic from the NATS implementation and keeps each contract scoped to what's actually used - smaller, consumer-owned interfaces are easier to reason about and mock than one large shared one.

`downlinks.Publisher`, `modbus.Publisher`, and `converters.Publisher` are currently identical - kept separate since each may need different capabilities as it evolves.

Resolves #1303

### DIFF
--- a/cmd/downlinks/main.go
+++ b/cmd/downlinks/main.go
@@ -272,7 +272,7 @@ func subscribeToThingsES(ctx context.Context, svc downlinks.Service, cfg config,
 	return subscriber.Subscribe(ctx, handler)
 }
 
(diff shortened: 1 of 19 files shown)


manuio (approved):
LGTM

PR #1307, opened 2026-08-19
MF-1204 - Clean up consumer handler types

### LINKED ISSUE #1204: Clean up consumer/message handler types and interfaces

majabirmancevic:
### Summary

- **Removed dead `UnsubscribeX` methods**: `AlarmSubscriber.UnsubscribeAlarms`, `CommandSubscriber.UnsubscribeCommands`, `NotificationSubscriber.UnsubscribeNotifications`, and `WebhookSubscriber.UnsubscribeWebhooks` were declared and implemented but never called anywhere - those consumers subscribe once at service startup and run for the process's lifetime, so nothing ever needs to unsubscribe them. Deleted the four unused methods.

- **Simplified typed handlers**: `AlarmHandler`/`CommandHandler`/`NotificationHandler`/`WebhookHandler` were `Handle`+`Cancel` interfaces, but none of their subscribers ever need cleanup - only `Subscribe` (Message) has live, per-connection callers (`ws`/`coap`) that do. Converted the four to plain func types (`func(subject string, x T) error`), matching how they're actually used.

- **Consequence in `consumers/consumer.go` and `mqtt/forwarder.go`**: the wrapper structs that existed solely to satisfy the old `Handle`+`Cancel` interfaces (`alarmHandler`, `notificationHandler`, `webhookHandler`, `handleFunc`) are gone - callers now pass a bound method or func directly (`c.ConsumeAlarm`, `c.ConsumeNotification`, etc.).

- **`pkg/messaging/nats/pubsub.go`**: collapsed four near-identical per-type NATS callback builders (`natsAlarmHandler`, `natsCommandHandler`, ... ) and their duplicated subscribe/dedup/locking bodies into one shared `subscribe`/`subscribeWithCancel` pair, driven by small `newFn`/`handleFn` closures built per `SubscribeX` method. 

- **Package split**: `pkg/messaging/pubsub.go` split into `publisher.go` and `subscriber.go`.

Closes #1204


pkg/messaging/publisher.go
```diff
@@ -0,0 +1,105 @@
...
+type Publisher interface {
+	// Publish publishes message to the message broker.
+	Publish(subject string, msg protomfx.Message) error
+
+	// Close gracefully closes message publisher's connection.
+	Close() error
+}
+
+// PubSub represents aggregation interface for publisher and subscriber.
+type PubSub interface {
```
majabirmancevic, line 36: `PubSub` in `pkg/messaging/publisher.go` is unused within this repo - its only remaining reference is `pkg/messaging/mqtt`'s `Subscribe`/`Unsubscribe`/`NewPubSub`, which has no production caller. The only feature that uses `pkg/messaging/mqtt` (the MQTT forwarder) only ever uses its `Publisher` side. 
Removing it is its own follow-up cleanup, intentionally left out of this PR to keep the diff scoped. Once that lands, `PubSub` should be deleted too.

duolok (approved):
LGTM

manuio (approved):
LGTM

PR #1309, opened 2026-08-21
MF-1308 - Add NoTotal field in PageMetadata

### LINKED ISSUE #1308: Make total count optional in reader message queries
`ListJSONMessages`/`ListSenMLMessages` always run a `SELECT COUNT(*) ... <same WHERE clause>` alongside the page query to fill `PageMetadata.Total`. Cost scales with matched rows, not the limit, and there's no way for callers to opt out.
Add an opt-out flag to `MessagesPageMetadata` that skips `COUNT(*)` in both repositories when the caller doesn't need the total.

majabirmancevic:
Resolves #1308

### DIFF
--- a/readers/timescale/senml.go
+++ b/readers/timescale/senml.go
@@ -129,9 +129,10 @@ func (sr *senmlRepository) readAll(ctx context.Context, rpm readers.SenMLPageMet
 			return page, err
 		}
 		page.Messages = messages
-		if !rpm.NoTotal {
-			page.Total = total
+		if rpm.NoTotal {
+			return page, nil
 		}
+		page.Total = total
 
 		return page, nil
 	}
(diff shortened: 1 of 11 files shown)


readers/timescale/senml.go
```diff
@@ -129,6 +129,12 @@ func (sr *senmlRepository) readAll(ctx context.Context, rpm readers.SenMLPageMet
 			return page, err
 		}
 		page.Messages = messages
+		if rpm.NoTotal {
+			// readAggregatedSenMLMessages already computes the total as part of the
+			// same round trip, over the already-bucketed rows.
+			// NoTotal just means we don't return that value via page.Total.
+			return page, nil
```
duolok: Do we need the comments?

majabirmancevic: There is a specific case where the query for `COUNT` is embedded in the main `SELECT` query, and regardless of `NoTotal` it will be executed, but will not be returned in the result (there is no point in separating the queries there), while with other methods, the `COUNT` query is executed or not executed depending on the `NoTotal` filter, so that's why I wanted to mention it in the comment

duolok (commented):
LGTM

manuio (approved):
LGTM

duolok (approved):
LGTM

PR #1315, opened 2026-08-28
MF-1311 - Support CSV/JSON upload to configure Modbus registers

### LINKED ISSUE #1311: Support CSV/JSON upload to bulk-configure Modbus registers
Register configuration is currently a manual, per-register process - for a client with many registers, this is slow and repetitive.

Support uploading a CSV or JSON file that defines a list of registers, and import it to configure all of them for a client in a single operation, instead of adding them one by one through the form.

majabirmancevic:
Resolves #1311

### DIFF
--- a/modbus/api/http/csv.go
+++ b/modbus/api/http/csv.go
@@ -13,7 +13,7 @@ import (
 	"github.com/MainfluxLabs/mainflux/pkg/errors"
 )
 
-// parseCSVFields mirrors the header-driven CSV convention converters/service.go uses for uploaded files.
+// parseCSVFields reads one register per row, matching columns to a header row case-insensitively.
 func parseCSVFields(r io.Reader) ([]field, error) {
 	reader := csv.NewReader(r)
 	reader.FieldsPerRecord = -1
--- a/modbus/api/http/csv.go
+++ b/modbus/api/http/csv.go
@@ -79,7 +79,8 @@ func parseCSVFields(r io.Reader) ([]field, error) {
 			if err != nil {
 				return nil, errors.New(fmt.Sprintf("line %d: invalid %s", line, colAddress))
 			}
-			f.Address = uint16(v)
+			addr := uint16(v)
+			f.Address = &addr
 		}
 
 		if s := get(row, colLength); s != "" {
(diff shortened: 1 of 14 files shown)


duolok (dismissed):
LGTM

modbus/api/http/csv.go
```diff
@@ -0,0 +1,87 @@
...
+	for i, name := range header {
+		col[strings.ToLower(strings.TrimSpace(name))] = i
+	}
+
+	get := func(row []string, name string) string {
+		i, ok := col[name]
+		if !ok || i >= len(row) {
+			return ""
+		}
+		return row[i]
```
duolok: This should be :

> return strings.TrimSpace(row[i])

modbus/api/http/csv.go
```diff
@@ -0,0 +1,87 @@
...
+func parseCSVFields(r io.Reader) ([]field, error) {
+	reader := csv.NewReader(r)
+	reader.FieldsPerRecord = -1
+
+	header, err := reader.Read()
+	if err == io.EOF {
+		return nil, nil
+	}
+	if err != nil {
+		return nil, errors.Wrap(ErrInvalidCSVRow, err)
```
manuio: Why `ErrInvalidCSVRow`?

duolok (changes requested):

modbus/api/http/csv.go
majabirmancevic: fixed

modbus/api/http/requests.go
```diff
@@ -27,6 +27,7 @@ var (
 	ErrInvalidFieldType    = errors.New("invalid field type")
 	ErrInvalidFieldLength  = errors.New("invalid field length")
 	ErrInvalidByteOrder    = errors.New("invalid byte order")
+	ErrInvalidCSVRow       = errors.New("invalid csv row")
```
manuio: `invalid CSV fields`?

manuio (changes requested):

modbus/api/http/csv.go
majabirmancevic: no need (it's removed)

modbus/api/http/requests.go
majabirmancevic: unnecessary error type; I removed it

manuio (approved):
LGTM

PR #1326, opened 2026-09-08
MF-1325 - Convert XML field to proper type

### LINKED ISSUE #1325: XML response values are all serialized as strings in downlinks
When a downlink returns XML, the parser keeps every value as text, so numbers and booleans end up quoted after conversion to JSON (e.g. `<temp>23.5</temp>` becomes `"temp":"23.5"`).
The fix is to infer scalar types during XML parsing.

majabirmancevic:
Resolves #1325


manuio (approved):
LGTM

duolok (approved):
LGTM

PR #1337, opened 2026-09-15
MF-1335 - Introduce Script as Rule condition

### LINKED ISSUE #1335: Introduce Script as Rule condition
Make a Lua script usable as a Rule **condition**: the script returns a boolean, which is combined with the rule's threshold conditions under its `Operator`. 

### Scope

- Add a `type` discriminator to `Condition` (`threshold` default / `script`) plus a `script_id`. Existing conditions (no `type`) keep behaving as threshold conditions.
- Evaluate `script` conditions inside the rule engine: load the referenced script, run it with a **read-only** Lua API (`log`, `list_messages` - no `smtp_notify` / `create_alarm`), and use its return value.
- Fail closed: a non-boolean return, no return, or a runtime error ⇒ condition is `false` and the rule continues.
- Keep the existing array-payload behaviour (evaluate per element, trigger if any element matches).
- Persist a `ScriptRun` for each evaluation.

### Validation

- `script` condition requires a valid `script_id` and must not carry `field` / `comparator` / `threshold`.
- The referenced script must exist and belong to the same group as the rule.

### Docs

- `rules/README.md` and `api/openapi/rules.yml`: document `type` / `script_id` on conditions, the read-only API, and fail-closed semantics.

majabirmancevic:
Resolves #1335


duolok (approved):
LGTM

manuio (approved):
LGTM

PR #1340, opened 2026-09-17
MF-1339 - Add list script runs by rule

### LINKED ISSUE #1339: Add list script runs by rule
`ScriptRun` records which script executed and which thing triggered it, but not which rule's condition caused the run - so there's no way to see a rule's script-condition history, only a thing's (via the existing `/things/:id/runs`). Add a `rule_id` column to `lua_script_runs` (nullable, historical rows predate it), thread the triggering rule's ID through condition evaluation, and expose `GET /rules/:id/runs` mirroring the existing thing-scoped endpoint (same status/from/to/name filters).

majabirmancevic:
Resolves #1339


duolok (dismissed):
LGTM

rules/postgres/scripts.go
```diff
@@ -288,11 +288,19 @@ func (rr ruleRepository) RemoveScriptRuns(ctx context.Context, ids ...string) er
...
 func (rr ruleRepository) RetrieveScriptRunsByThing(ctx context.Context, thingID string, pm rules.PageMetadata) (rules.ScriptRunsPage, error) {
+	return rr.retrieveScriptRuns(ctx, "thing_id", thingID, pm)
+}
+
+func (rr ruleRepository) RetrieveScriptRunsByRule(ctx context.Context, ruleID string, pm rules.PageMetadata) (rules.ScriptRunsPage, error) {
+	return rr.retrieveScriptRuns(ctx, "rule_id", ruleID, pm)
+}
+
+func (rr ruleRepository) retrieveScriptRuns(ctx context.Context, col, val string, pm rules.PageMetadata) (rules.ScriptRunsPage, error) {
 	oq := dbutil.GetOrderQuery(pm.Order, rules.RuleOrderFields)
```
manuio: Create the equivalent of `rules.RuleOrderFields` for script runs

manuio (changes requested):

rules/postgres/scripts.go
majabirmancevic: fixed

manuio (approved):
LGTM

duolok (approved):
LGTM

